PR #294, opened 2014-12-22
Refactor video handling in stream

caotto:
Introduce gallery subclasses for reading videos, and use those instead of more specialized classes in stream.

Also, handle webcams (opened by index) similarly to webcamFormat (i.e. accept them as 0.webcam, or whatever) to avoid ambiguity in how to open files.

There are multiple commonly used video extensions, and we already have a DefaultGallery, so this PR explicitly enumerates some extensions with subclasses of the common videoGallery. This is perhaps not ideal in general, but no better ideas so far.


bklare:
@caotto Do you see anything wrong the command below? 

br -gui -algorithm "Stream(Open+Show+Discard)" -enroll 0.webcam
> Set algorithm to Stream(Open+Show+Discard)
> Enrolling 0.webcam
> WARNING: Couldn't read movie file /Users/m29389/Source/OpenBR/closedbr/0.webcam
> stream failed to open 0.webcam
> 100.00%  ELAPSED=00:00:00  REMAINING=00:00:00  COUNT=1

caotto:
Yes, basically if you do
br -gui -algorithm "Show(false)" -enroll 0.webcam
it should work. This leverages the stream that is implicitly used in
enroll, and reads frames from the webcam like a gallery.

If you use an explicit stream (as in your example), then the situation gets
complicated, because the implicit stream will read frames from the webcam
like a gallery, then the explicit stream will try to open every frame read
by the implicit stream. This will lead to problems eventually no matter
what (although the one that actually surfaced is in fact a little odd).

Basically you can think about it like this, enrolling from 0.webcam in this
context uses the webcam as a gallery, and the stream reads frames from the
webcam similar to reading lines from a .txt or .xml file. If you use an
explicit stream, it won't work properly, unless you do something like put
"0.webcam" as one line of a .txt file, and enroll the .txt file (in this
case, the 0.webcam line will be read by the implicit stream, and
distributed to the explicit one appropriately).

PR #303, opened 2015-01-07
Update Classification API to support OpenBR object handling

JordanCheney:
This should take care of the overhead to create the Representation and Classification abstractions in OpenBR. @jklontz @caotto could one of you review when you have time?


jklontz:
This LGTM, perhaps @caotto can review also?

PR #304, opened 2015-01-08
try masking the impostor scatter matrix

jklontz:
@bklare I would appreciate if you could review this code for conceptual correctness. This is the implementation of the idea floated yesterday to try sampling the between-class scatter matrix to include only difficult examples. I've tried various weighting strategies based on the L2 distance between class means, including the binary "mask" version shown here. Unfortunately, none of them performed better than keeping the entire, unmodified, between-class scatter matrix.

### DIFF
--- a/openbr/plugins/eigen3.cpp
+++ b/openbr/plugins/eigen3.cpp
@@ -130,11 +130,11 @@ private:
     }
 
 protected:
-    void trainCore(Eigen::MatrixXd data)
+    void trainCore(Eigen::MatrixXd data, Eigen::MatrixXi mask = Eigen::MatrixXi())
     {
         int dimsIn = data.rows();
         int instances = data.cols();
-        const bool dominantEigenEstimation = (dimsIn > instances);
+        const bool dominantEigenEstimation = (dimsIn > instances) || ((mask.rows() == instances) && (mask.cols() == instances));
 
         Eigen::MatrixXd allEVals, allEVecs;
         if (keep != 0) {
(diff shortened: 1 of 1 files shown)


openbr/plugins/eigen3.cpp
```diff
@@ -454,6 +459,16 @@ class LDATransform : public Transform
             for (int i=0; i<space1.keep; i++) space1.eVecs.col(i) /= pow((double)space1.eVals(i),0.15);
         }
 
+        Eigen::MatrixXi betweenClassMask;
+        if (betweenClassThreshold != 0) {
+            Eigen::MatrixXd distances(numClasses, numClasses);
+            for (int i=0; i<numClasses; i++)
+                for (int j=0; j<numClasses; j++)
+                    distances(i, j) = (classMeans.col(i) - classMeans.col(j)).squaredNorm();
+            betweenClassMask = (distances.array() < (betweenClassThreshold * distances.mean())).cast<int>();
```
bklare: @jklontz How much did you play around with the threshold here?

jklontz: A lot. I tried masking out from ~3% through ~97% of the data.

openbr/plugins/eigen3.cpp
```diff
@@ -148,6 +148,9 @@ class PCATransform : public Transform
             if (dominantEigenEstimation) cov = data.transpose() * data / (instances-1.0);
             else                         cov = data * data.transpose() / (instances-1.0);
 
+            if ((mask.rows() == cov.rows()) && (mask.cols() == cov.cols()))
+                cov = cov.array() * mask.cast<double>().array();
```
bklare: @jklontz The covariance matrix represents feature dimensions, not instances. If I undestand this correctly, you seem to be masking out features in the covariance matrix. Instead, I think you need to compute the covaraince matrix at the instances level, skipping over masked instances. This would mean not computing the full covariance matrix in the manner done on lines 148-149, but by doing the matrix multipliactions one at a time depending on whether the instance pairs are being used or not.  After that, the covariance matrix should be usable in the same manner, the only difference being that it will not have computed the covariance between feature vectors that are highly separated already

jklontz: @bklare That's what I had originally planned on doing. But isn't the current implementation correct when `dominantEigenEstimation = true`, which is the case in this circumstance, and `cov` has size `instances` X `instances`.

bklare: @jklontz I see. Yes, in the case where domEig is true then my concern is not valid. I am not sure, though, if this domEig method holds with this instance removal approach come line 158 (would have to think about this). It probably does, but if so motivated might be worth some pencil and paper to be sure.

PR #305, opened 2015-01-12
introduced br::Distance::trainable() and UntrainableDistance

jklontz:
@caotto @sklum @bklare The following PR introduces `virtual bool br::Distance::trainable()` and `br::UntrainableDistance`, the distance helper class analogy to `br::UntrainableTransform`. The purpose of this PR is to avoid projecting the training set through the enrollment pipeline, only to pass it to a distance that doesn't require any training. This is a substantial time saver for algorithms with untrainable distance metrics and large training sets.

### DIFF
--- a/openbr/core/core.cpp
+++ b/openbr/core/core.cpp
@@ -77,7 +77,7 @@ struct AlgorithmCore
         qDebug("Training Enrollment");
         trainingWrapper->train(data);
 
-        if (!distance.isNull()) {
+        if (!distance.isNull() && distance->trainable()) {
             if (Globals->crossValidate > 0)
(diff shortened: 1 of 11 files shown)


openbr/core/core.cpp
```diff
@@ -77,7 +77,7 @@ struct AlgorithmCore
         qDebug("Training Enrollment");
         trainingWrapper->train(data);
 
-        if (!distance.isNull()) {
+        if (!distance.isNull() && distance->trainable()) {
```
jklontz: Here's the important optimization.

caotto:
:+1: long awaited

bklare:
Excellent! If only we could go back and get all that time that was wasted on this computation before...

sklum:
Thanks for this @jklontz.

PR #306, opened 2015-01-13
break ties pessimistically

jklontz:
@bklare @sklum @bhklein As described.

### DIFF
--- a/openbr/core/eval.cpp
+++ b/openbr/core/eval.cpp
@@ -36,7 +36,12 @@ struct Comparison
     Comparison() {}
     Comparison(float _score, int _target, int _query, bool _genuine)
         : score(_score), target(_target), query(_query), genuine(_genuine) {}
-    inline bool operator<(const Comparison &other) const { return score > other.score; }
+
+    inline bool operator<(const Comparison &other) const
+    {
+        if (score != other.score) return (score > other.score);
+        else                      return !genuine; // Tie-break favors pessimistic behavior of ranking impostors higher.
+    }
 };
 
 #undef FAR // Windows preprecessor definition conflicts with variable name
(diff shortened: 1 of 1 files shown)


openbr/core/eval.cpp
```diff
@@ -36,7 +36,12 @@ struct Comparison
     Comparison() {}
     Comparison(float _score, int _target, int _query, bool _genuine)
         : score(_score), target(_target), query(_query), genuine(_genuine) {}
-    inline bool operator<(const Comparison &other) const { return score > other.score; }
+
+    inline bool operator<(const Comparison &other) const
+    {
+        if (score != other.score) return (score > other.score);
+        else                      return !genuine; // Tie-break favors pessimistic behavior of ranking impostors higher.
```
caotto: @jklontz @bhklein 
Interestingly using msvc2012, in debug mode, the following debug error is triggering:

template<class _Ty1, class _Ty2> inline
	bool _Debug_lt(_Ty1& _Left, _Ty2& _Right,
		_Dbfile_t _File, _Dbline_t _Line)
	{	// test if _Left < _Right and operator< is strict weak ordering
	if (!(_Left < _Right))
		return (false);
	else if (_Right < _Left)
		_DEBUG_ERROR2("invalid operator<", _File, _Line);
	return (true);
	}

With this operator, if the left and right are impostors with equal score, then 
both L < R and R < L are true (so the operator is indeed not well formed). If you use
return !genuine && other.genuine; 
this is avoided.

jklontz:
Good point, I think Ben ran into this issue and had to use stable sort. Will push the fix.

jklontz:
https://github.com/biometrics/openbr/commit/922d0217a5d85cb52fe1705ec5edfb56cce2a910

PR #307, opened 2015-01-14
rewrote StopWatchTransform

jklontz:
@JordanCheney This proposed rewrite of StopWatchTransform adds the following features:
1. Holds on to the transform description so that it can be printed, making it sensible to use multiple instantiations of StopWatch in the same algorithm.
2. Works with trainable transforms.
3. No longer locks on the child->project call, allowing for better parallelization.

Let me know if you think this introduces any regressions or is otherwise too invasive. Happy to tweak it or rename it as a separate transform.


JordanCheney:
LGTM. The functionality for the old transform was pretty limited.

PR #309, opened 2015-01-14
Remove memGallery::align and associated infrastructure

caotto:
align didn't appear to improve speed, and allocating a single array for all template data is not feasible for large enough galleries


jklontz:
LGTM

Issue #302, opened 2015-01-05
Improved evalLandmarking/plotLandmarking functionality


sklum:
It would be nice if `plotLandmarking` incorporated some of the new functionality present in `plot`. Specifically, I'd like the PDF to contain:

* A single example with the points drawn and labelled 0-n on the face.
* Tables with mean error, median error, and std. dev. of error for each landmark along with an with the average of these values over all landmarks.
* Mean normalization distance (in pixels)
* Most accurate/least accurate examples.  Here, I'd like images side by side with the ground truth drawn on the left (without labels) and automatically determined on the right (without labels).

sklum:
@bhklein As most of the computation for this is already present in `evalLandmarking`, it's simply a matter of translating things into the proper R calls.  Given your expertise with the `eval` framework and R, do you mind looking into this?  I want to emphasize that this is _not_ a priority.

bhklein:
Sure, I can work on implementing this sometime in the coming weeks.

PR #310, opened 2015-01-15
Eval landmark improvements

sklum:
This branch introduces functionality to evalLandmarking including:

* A sample image with ground truth landmarks drawn and labeled with their index that corresponds to later figures.
* Accurate and inaccurate testing examples.
* Table of average error rates for each landmark with confidence intervals.
* Reordered results in the normalized error plots such that they are sequential, as opposed to ordered by factor.
* Added options to draw to control the display of landmarks when using the `verbose` flag and to optionally display named points.

There is some duplication here with respect to the R code, which I can eliminate if this bothers anyone.

@bhklein @jklontz take a look?

Example posted in Slack.

Solves #302.

### DIFF
--- a/app/br/br.cpp
+++ b/app/br/br.cpp
@@ -163,8 +163,8 @@ public:
                 check((parc >= 2) && (parc <= 6), "Incorrect parameter count for 'evalDetection'.");
                 br_eval_detection(parv[0], parv[1], parc >= 3 ? parv[2] : "", parc >= 4 ? atoi(parv[3]) : 0, parc >= 5 ? atoi(parv[4]) : 0, parc == 6 ? atoi(parv[5]) : 0);
             } else if (!strcmp(fun, "evalLandmarking")) {
-                check((parc >= 2) && (parc <= 5), "Incorrect parameter count for 'evalLandmarking'.");
-                br_eval_landmarking(parv[0], parv[1], parc >= 3 ? parv[2] : "", parc >= 4 ? atoi(parv[3]) : 0, parc >= 5 ? atoi(parv[4]) : 1);
+                check((parc >= 2) && (parc <= 6), "Incorrect parameter count for 'evalLandmarking'.");
+                br_eval_landmarking(parv[0], parv[1], parc >= 3 ? parv[2] : "", parc >= 4 ? atoi(parv[3]) : 0, parc >= 5 ? atoi(parv[4]) : 1,  parc >= 6 ? atoi(parv[5]) : 0);
             } else if (!strcmp(fun, "evalRegression")) {
                 check(parc >= 2 && parc <= 4, "Incorrect parameter count for 'evalRegression'.");
                 br_eval_regression(parv[0], parv[1], parc >= 3 ? parv[2] : "", parc >= 4 ? parv[3] : "");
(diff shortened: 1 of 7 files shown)


app/br/br.cpp
```diff
@@ -163,8 +163,8 @@ class FakeMain : public QRunnable
                 check((parc >= 2) && (parc <= 6), "Incorrect parameter count for 'evalDetection'.");
                 br_eval_detection(parv[0], parv[1], parc >= 3 ? parv[2] : "", parc >= 4 ? atoi(parv[3]) : 0, parc >= 5 ? atoi(parv[4]) : 0, parc == 6 ? atoi(parv[5]) : 0);
             } else if (!strcmp(fun, "evalLandmarking")) {
-                check((parc >= 2) && (parc <= 5), "Incorrect parameter count for 'evalLandmarking'.");
-                br_eval_landmarking(parv[0], parv[1], parc >= 3 ? parv[2] : "", parc >= 4 ? atoi(parv[3]) : 0, parc >= 5 ? atoi(parv[4]) : 1);
+                check((parc >= 2) && (parc <= 7), "Incorrect parameter count for 'evalLandmarking'.");
```
jklontz: update br -help also

openbr/core/eval.cpp
```diff
@@ -1049,39 +1059,94 @@ float EvalLandmarking(const QString &predictedGallery, const QString &truthGalle
...
 
     QStringList lines;
     lines.append("Plot,X,Y");
+
+    QtUtils::touchDir(QDir("landmarking_examples_truth"));
+    QtUtils::touchDir(QDir("landmarking_examples_predicted"));
+
+    // Example
+    {
+        QScopedPointer<Transform> t(Transform::make("Open+Draw(verbose,rects=false,location=false)",NULL));
```
jklontz: I don't like the fact that we have to construct and project through a transform. Is it that hard to render the data offline and then point to it with `-path`, like we do in `-eval`? This would allow better orthogonality of function.

jklontz:
If there is work to be done cleaning up the R code, I think better to do it now before you move on to something else :)

jklontz:
@bhklein Please review

openbr/core/eval.cpp
sklum: I agree it's not very elegant but is rendering all of the images when we only need a small subset of them just so we don't have to make a transform an improvement?

bhklein: I think the landmark plotting can all be done in R by using the image as the plot background and overlaying the points that way R can just use the original image and there's no need for writing any more to disk.

bhklein:
The example output looks good to me, I can work on consolidating some of the R code and the way the landmarks are plotted that I mentioned above sometime this weekend.

openbr/core/eval.cpp
sklum: In theory, yes, this is probably the best way to do it.  It's a bit tricky in that you'll need to convert between the image coordinates and the graph coordinates.  I decided to follow the adage of "perfect is the enemy of good" when implementing the annotation, but I can implement it this way now that I understand R a bit better if that's what is desired.

sklum:
Fantastic! Thanks @bhklein.

bhklein:
Sorry I finally got around to taking a look at this a few days ago and forgot to post an update.  The two approaches I tried were scaling the points over an annotation_custom of the raster, and actually plotting the image's pixels.  I haven't been quite able to get the scaling right for the former and the latter was <em>extremely</em> slow so I don't see it is an option.  I'm fine with merging it as it is if Josh is okay with calling project in eval, otherwise we could still easily change that to the offline method.  I'm going to take a last look at doing this in R at some point but I'm not sure about how feasible it is.

jklontz:
I'm ok with merging

PR #311, opened 2015-01-16
Updates to stored transform handling

caotto:
2 parts here, one is a refactor of setPropertyRecursive, so that the implementation is a little more general. Instead of implementing a method in every Transform with child transforms, the default implementation of setPropertyRecursive looks for Transform * and QList<Transform *> properties automatically. 

There are some limitations to this approach, in particular all transforms which use a QString property to store their child transforms description, rather than storing the child transform as a property still require special casing. It is perhaps worth reconsidering this idiom in general, since by this point using a QString property loses out on both automatic serialization of the child transform, and proper navigation of transform trees via reflection. 

Part 2, is as requested by @sklum, adding support for additional arguments to transforms stored via LoadStore (though the implementation is perhaps more aggressive than anticipated).

Per 9abe69538:
If an unknown transform string is encountered in an algorithm, attempt
to load it via LoadStore, e.g.
"Open+Cascade+landmark_model"
would expand landmark_model to <landmark_model>. Additionally, arguments
supplied to such unknown transforms are set via setPropertyRecursive
e.g.
"Open+Cascade+landmark_model(something=true)"
works.

Also allow _ in model names specified as the first argument to LoadStore

### DIFF
--- a/openbr/openbr_plugin.cpp
+++ b/openbr/openbr_plugin.cpp
@@ -1413,6 +1413,13 @@ Transform *Transform::make(QString str, QObject *parent)
     if (str.startsWith('(') && str.endsWith(')'))
         return make(str.mid(1, str.size()-2), parent);
 
+    // Base name not found? Try constructing it via LoadStore
+    if (!Factory<Transform>::names().contains(parsed.suffix())) {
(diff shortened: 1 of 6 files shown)


jklontz:
LGTM

PR #312, opened 2015-01-22
Adding a minimum version requirement for OpenCV (2.4.5)

stephenrawls:
The build was failing for me because I have multiple versions of OpenCV installed on my system, and the CMake build was just using whatever the default OpenCV was, which in my case was an older version that didn't support some of the cascade API OpenBR was using.

The solution is to give the find_package command a minimum version requirement for OpenCV. I chose version 2.4.5 because that is documented at the following website: http://openbiometrics.org/doxygen/latest/linux_gcc.html


jklontz:
Thanks

PR #315, opened 2015-01-28
Implemented RndSample transform

sklum:
Randomly samples around a specified point to provide positive and negative samples to a classifier/regressor.


jklontz:
lgtm

PR #319, opened 2015-01-30
Use a uint8 lookup table for the gamma filter

caotto:
The opencv function LUT expects a uint8 lookup table, using a float leads to unusual results (low gamma becomes brighter for whatever reason).

Fixing this does invalidate any models trained using the original version (FaceRecognition etc.)

### DIFF
--- a/openbr/plugins/filter.cpp
+++ b/openbr/plugins/filter.cpp
@@ -38,15 +38,26 @@ class GammaTransform : public UntrainableTransform
 
     void init()
     {
-        lut.create(256, 1, CV_32FC1);
-        if (gamma == 0) for (int i=0; i<256; i++) lut.at<float>(i,0) = log((float)i);
-        else            for (int i=0; i<256; i++) lut.at<float>(i,0) = pow(i, gamma);
+        lut.create(256, 1, CV_8UC1);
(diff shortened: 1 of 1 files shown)


caotto:
@dgcrouse fyi, this fix cannot be merged yet, since at the very least the pre-trained openbr models need to be retrained, but you can use this branch in isolation if you want

caotto:
@dgcrouse fyi, this fix cannot be merged yet, since it invalidates the current pre-trained openbr models, but you can use it in isolation if you want.

jklontz:
This seems to warrant issuing a bug with OpenCV

caotto:
gpu::LUT is commented with this information:
http://docs.opencv.org/modules/gpu/doc/operations_on_matrices.html#gpu-lut
although regular LUT is not:
http://docs.opencv.org/modules/core/doc/operations_on_arrays.html#lut
I'd guess just a documentation bug if anything

caotto:
Correction, using LUT (but not gpu::LUT, probably), converts the datatype of the matrix to float, which causes issues with displaying the image (since it gets rescaled to 0-1 from (truncated), 0-255) so the output is floating point values 14.2 or w/e rather than uint8 14

PR #320, opened 2015-02-02
Block based model compression scheme

caotto:
Avoid serialization failures due to uncompressed model size exceeding 2GB. The cause of this issue is that we serialize to an intermediate buffer before writing to disk, and the QByteArray used for this buffer cannot have more than INT_MAX elements. 

This branch avoids the issue by introduce a QIODevice subclass which buffers stream output into fixed size blocks, and compresses those individually rather than using a single buffer for the entire thing. 

This constitutes a binary compatibility break, so all pre-trained models have to be retrained. This is not done yet (FaceRecognition training is done, but gender classification/age regression have not been retrained yet). In spite of this, the code is more or less complete. 

Also, some additional updates to training, geared towards improving training speed and providing more informative output during training.

### DIFF
--- a/openbr/plugins/meta.cpp
+++ b/openbr/plugins/meta.cpp
@@ -96,17 +96,15 @@ class PipeTransform : public CompositeTransform
 
         int i = 0;
         while (i < transforms.size()) {
-            fprintf(stderr, "\n%s", qPrintable(transforms[i]->objectName()));
-
             // Conditional statement covers likely case that first transform is untrainable
             if (transforms[i]->trainable) {
-                fprintf(stderr, " training...");
+                qDebug() << "Training " << transforms[i]->description() << "\n...";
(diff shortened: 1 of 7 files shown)


openbr/plugins/meta.cpp
```diff
@@ -130,7 +130,16 @@ class PipeTransform : public CompositeTransform
                    !transforms[nextTrainableTransform]->timeVarying())
                 nextTrainableTransform++;
 
-            fprintf(stderr, " projecting...");
+            // No more trainable transforms? Don't need any more projects then
+            if (nextTrainableTransform == transforms.size())
```
jklontz: :)

jklontz:
lgtm

bklare:
Thanks, Charles.

PR #326, opened 2015-02-07
Gray the same in Read and Cvt

bklare:
Suggested change b/c it is annoying to have `Cvt(Gray)` and `Read(Grayscale)`

### DIFF
--- a/openbr/plugins/misc.cpp
+++ b/openbr/plugins/misc.cpp
@@ -103,7 +103,7 @@ public:
     enum Mode
     {
         Unchanged = IMREAD_UNCHANGED,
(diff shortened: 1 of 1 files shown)


jklontz:
ok, but please make sure our closedbr is updated accordingly after the merge.

jklontz:
Closed due to lack of interest.

bklare:
Why no merge?

bklare:
It does not make sense to have two functions doing effectively the same thing with different parameter names.

jklontz:
Agreed, but the commit doesn't merge cleanly, so if you want to re-submit with corrections in algorithms.cpp etc. then we can apply it.

PR #334, opened 2015-02-19
Explicit bytearray serialization

caotto:
When serializing bytearrays in BlockCompression, explicitly use read/writeRawData instead of the default QDataStream serialization. QDataStream allegedly serializes byte arrays as quint32 length, followed by raw bytes, but in practice serialization failures occur in some circumstances (typically saving very large model files, involving multiple CvStatModel serializations), and are avoided (to my knowledge) by doing the documented serialization explicitly ourselves. 

This does not break binary compatibility, since it just explicitly does what the documentation claims QDataStream does.


jklontz:
lgtm

PR #335, opened 2015-02-19
Remove dependency on QT5Svg

caotto:
As far as I know, nothing is using this, and it's not included in the qt5 debian package on one of our servers


jklontz:
I think that QtSvg was used at some point for icons when developing GUIs. I'm ok with it going away, but @sklum may need it conditionally enabled.

PR #336, opened 2015-02-24
Add a simple binary .ut format, add matlab scripts to write to it

caotto:


jklontz:
I would appreciate this being generalized to support channels also, that's the only thing missing from being able to specify arbitrary OpenCV matrices in the .ut format.

jklontz:
lgtm

PR #337, opened 2015-02-25
ReadKeyframes transform and libav videocapture class

bhklein:
This introduces the `ReadKeyframes` transform, effectively replacing the need to call `ffmpeg` as an offline step to extract key frames.  The transform outputs key frame templates with time stamps and frame numbers.

@jklontz or anyone else interested please review


caotto:
Why a transform, not a gallery for reading the frames?

caotto:
To elaborate, I'd say AVCapture is certainly redundant with the Gallery class hierarchy, and from that perspective, ReadKeyFrames is redundant with a Transform that reads from Gallery.

I also question the wisdom of reading all frames, then processing them in memory at once (in general), it's fine for short videos, but you'll regret it sooner or later.

bhklein:
Good points.  What are your thoughts on leaving AVCapture as is, and using it in the same way cv::VideoCapture is used currently in video galleries in a new keyframe gallery?  That way it can be still be used outside of the context of just galleries.

caotto:
I mean, galleries are surfaced better than a random class in some file due
to the general lack of header files in this project--so from that
perspective galleries are better in terms of ease of re-use.

On the other hand I don't think it's very important either way if there is
a KeypointGallery class that is a shallow interface to AVCapture vs. having
a single class

bhklein:
Okay makes sense, I'll go ahead and change it to a gallery, thanks.

bhklein:
@caotto here's the gallery implementation

openbr/plugins/gallery/keyframes.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    ~keyframesGallery()
+    {
+        release();
+    }
+
+    virtual void deferredInit()
+    {
+        av_register_all();
+        avformat_network_init();
```
caotto: I'm not familiar with libav, is it appropriate for these two functions to be called more than once per execution of the application? If yes they're fine here, otherwise move these calls to a separate Initializer subclass

openbr/plugins/gallery/keyframes.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+            QString URL = file.get<QString>("URL", file.name);
+            output.file.set("URL", URL + "#t=" + QString::number((int)(idx * time_base)) + "s");
+            output.file.set("frame", QString::number(idx * time_base * fps));
+            TemplateList dst;
+            dst.append(output);
+            return dst;
+        }
+        *done = true;
+        return TemplateList();
```
caotto: I don't think this is correct

openbr/plugins/gallery/keyframes.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    void release()
+    {
+        if (avSwsCtx)     sws_freeContext(avSwsCtx);
+        if (frame)        av_free(frame);
+        if (cvt_frame)    av_free(cvt_frame);
+        if (avCodecCtx)   avcodec_close(avCodecCtx);
+        if (avFormatCtx)  avformat_close_input(&avFormatCtx);
+        if (buffer)       av_free(buffer);
+    }
```
caotto: These pointers aren't guaranteed to be initialized to null, so in a case where deferredInit is not called, this is liable to crash. You should initialize them to NULL in the constructor, and if release can be called more than once, you should also NULL the pointers here so that you don't double-free them in that condition.

openbr/plugins/gallery/keyframes.cpp
bhklein: These functions have a static flag tracking whether they've been called yet, so I don't think this should be a problem.  I've also called them more than once in my testing and haven't ran into any problems.

openbr/plugins/gallery/keyframes.cpp
bhklein: good catch

caotto:
OK looks fine by me

bhklein:
great thanks

PR #339, opened 2015-03-02
Added new transform

bklare:
A subset of the [previous pull request](https://github.com/biometrics/openbr/compare/averagePoints) (sans eye drawing). This is a new pull request since the previous pull request is not valid after recent refactor.

### DIFF
--- a/openbr/plugins/metadata/averagepoints.cpp
+++ b/openbr/plugins/metadata/averagepoints.cpp
@@ -1,0 +1,52 @@
+#include <openbr/plugins/openbr_internal.h>
+
+namespace br
+{
+
+/*!
+ * \ingroup transforms
+ * \brief Averages a set of landmarks into a new landmark
+ * \author Brendan Klare \cite bklare
+ */
+class AveragePointsTransform : public UntrainableMetadataTransform
+{
+    Q_OBJECT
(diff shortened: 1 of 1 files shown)


jklontz:
LGTM, your code quality has really improved :)

openbr/plugins/metadata/averagepoints.cpp
```diff
@@ -0,0 +1,52 @@
...
+        for (int i = 0; i < indices.size(); i++) {
+            x1 += src.points()[indices[i]].x();
+            y1 += src.points()[indices[i]].y();
+        }
+
+        QPointF p(x1 / indices.size(), y1 / indices.size());
+        if (!metaName.isEmpty())
+            dst.set(metaName, p);
+        if (append)
+            dst.points().append(p);
```
sklum: As a heads up @bklare, `br::File` has the convenience methods `appendPoint(QPointF)` and `appendPoints(QPointF)`.

jklontz: Good observation, in fact this code doesn't work as written, `points()` returns a copy, not a reference.

bklare:
@sklum  @jklontz Thanks for the catch. Clearly this was an option I wasn't using. Fixed. Will merge now.

PR #342, opened 2015-03-03
Crop from mask 

bklare:
Crops an image based on an image metadata Mask


sklum:
Looks good!

bklare:
TY. Merging.

PR #343, opened 2015-03-06
Decompose k-NN graph generation and clustering

caotto:
separating k-NN graph computation and clustering is helpful in various ways. Our current br -cluster behavior is unchanged, and implementing new interfaces is now possible.

### DIFF
--- a/openbr/core/cluster.h
+++ b/openbr/core/cluster.h
@@ -22,16 +22,43 @@
 #include <QStringList>
 #include <QVector>
 #include <openbr/openbr_plugin.h>
+#include <openbr/plugins/openbr_internal.h>
 
 namespace br
 {
     typedef QList<int> Cluster; // List of indices into galleries
     typedef QVector<Cluster> Clusters;
 
-    Clusters ClusterGallery(const QList<cv::Mat> &simmats, float aggressiveness);
-    Clusters ClusterGallery(const QStringList &simmats, float aggressiveness, const QString &csv);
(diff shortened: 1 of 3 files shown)


openbr/core/cluster.h
```diff
@@ -22,16 +22,43 @@
...
+    // direct serialization to file system.
+    void  knnFromGallery(const QString &galleryName, const QString & outFile, int k = 20);
+    // in memory graph computation
+    Neighborhood knnFromGallery(const QString &gallery, int k = 20);
+
+    // Load k-NN graph from a file
+    Neighborhood loadkNN(const QString &fname);
+
+    // Save k-NN graph to file
+    bool savekNN(const Neighborhood &neighborhood, const QString &outfile);
```
jklontz: Please document the kNN file format

jklontz:
Thanks for doing this. My primary is whether or not the KNN file format would be better as a binary one than as the current text format. I'd like for what emerges to be our de-facto standard moving forward, and I want to make sure it scales to N=1e9. LGTM otherwise.

caotto:
sure, the current format is not very efficient. I'm not really planning on writing a binary serialization method in the immediate future, but it's something that should happen at some point. Generally, it can also be done as serializing a sparse adjacency matrix (instead of serializing edge lists as is currently done), so the best approach may be to adapt some sparse matrix format--but that requires further research

jklontz:
lgtm

Issue #344, opened 2015-03-06
load/store composition issue 


caotto:
We have run into an issue related to composing a pre-trained feature extractor and an SVM.
Basically, a procedure of the form:
br -algorithm "<...,FeatureExtraction>" -train train_data
br -algorithm "<FeatureExtraction>+SVM" -train other_train_data aModel
br -algorithm aModel -enroll test_data
results in what initially appears to be a serialization failure of FeatureExtraction in the final enroll, while running
br -algorithm "<FeatureExtraction>" -enroll data
on its own works.

caotto:
addressed via #345

PR #345, opened 2015-03-07
Look for sub-models in directories algorithms were found in 

caotto:
This commit is re: #344.
Consider a case where a partial model is trained, and serialized using load/store, then the complete algorithm is trained using that submodel, and saved as a complete algorithm (the output of -train):
br -algorithm "<Submodel>+OtherThings... " -train data finalModel
if the combined model is subsequently loaded, by specifying a relative or absolute path (i.e. not just loaded from the same directory), the submodel will not be found (since it is referred to in finalModel by its local name). i.e. 
br -algorithm /path/to/finalModel -enroll something
will not work.

This commit addresses that issue by also searching for sub-models in the paths were algorithms are successfully loaded from

### DIFF
--- a/openbr/core/core.cpp
+++ b/openbr/core/core.cpp
@@ -138,6 +138,16 @@ struct AlgorithmCore
 
     void load(const QString &model)
     {
+        // since we are loading an existing model, add its path to the search list for submodels
+        // assuming it is not already present.
+        QFileInfo finfo(model);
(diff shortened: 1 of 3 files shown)


jklontz:
I think that `modelSearch` is used frequently enough to warrant adding it as a field to the Context struct. This would simplify the PR a lot.

jklontz:
So much better :) lgtm

PR #349, opened 2015-03-11
Better stasm inclusion/compilation

sklum:
It ended up being way easier to fix some of the long standing problems related to `stasm`.  Thus, this branch:
* Simplifies how `stasm` is built into `openbr` by just including all the source.  To some extent this was happening before but it wasn't correctly done.
* Fixes the `Stasm_DIR` issue where cmake was finding an incorrect path.


jklontz:
lgtm

PR #353, opened 2015-04-02
Emd distance

sklum:
This branch introduces Earth Mover's Distance.


jklontz:
Did the OpenCV implementation not work as expected?

jklontz:
(lgtm)

sklum:
This is the OpenCV implementation and it works - it's very, very slow though.

PR #354, opened 2015-04-06
Docs

JordanCheney:
## Version 0.1 of the new Docs

Welcome to the new docs! This is version 0.1 because there is still a ton to do. I have now completed a little bit of everything and wanted to open up the new format and content to questions, comments, concerns etc.

## Viewing the docs

I am using a Github project called MkDocs to make these docs. To view them as a website you have to download the project and set it up. Alternatively, you can just open up the files since everything is now written in Markdown and renders nicely on Github. To install MkDocs docs open up terminal and type-
```
git clone https://github.com/mkdocs/mkdocs.git
cd mkdocs
python setup.py build
python setup.py install
```

Then to view the docs
```
cd /path/to/openbr/
cd docs
mkdocs serve
```

Then open up an internet browser and put <tt>http://127.0.0.1:8000/</tt> in the address bar and you should be all set.

## Completed

* Quick Start Tutorial: A new tutorial introducing users to the command line, plugins, and data structures using the webcam and face detection.

## ToDo

* Add at least one more tutorial about training algorithms in OpenBR. Open to other tutorial ideas
* Move face recognition, gender estimation, and age estimation tutorials to markdown
* Move the C API to markdown
* Move the command line API to markdown
* Add examples to the C++ Plugin API (~50% completed)
* Document properties in all plugins (classification folder is done as an example)

## Known Issues

* Code renders differently in different parts of the docs. This issue might be upstream.


jklontz:
So I feel like this PR presents two components that ought to be considered independently.

The first is the new content (API examples, property documentation, etc.) which is golden and can be committed directly to master without review using the existing documentation syntax (doxygen). Thank you for adding this stuff!

The second is a transition to a new documentation framework which IMO represents a lot of code churn for very little justified benefit. 
- Is the markdown rendering so much better that we move our API documentation from inline doxygen comments to separate markdown files?
- Are we liable to run into fewer problems with mkdocs, even though doxygen is more mature and specifically designed for C/C++? 
- It seems harder to maintain the documentation through refactoring under the proposed system, especially as it doesn't appear to offer any warnings when links become invalidated?
- What happens to the `\cite` command?
- Would the search functionality offered by doxygen be lost?
- Do we lose the option of rendering our documentation to tex?
- If the primary concern is that the doxygen generated HTML looks poor, why not use a different stylesheet?

Curious to hear your opinions @sklum @caotto. I feel like mkdocs represents a step backwards. Perhaps @JordanCheney elaborate on why it's better and worth the migration?

JordanCheney:
My thoughts-

* I think markdown rendering is cleaner in general then inline doxygen comments. IMO comments in code should be very specific and low-level and adding header-level comments (see the doxygen comment above br::File as an example) hurts readability. If I want general knowledge I would go to the project documentation. If I am poking around in the source code that means I want implementation specific details and digging through lines and lines of general information actually makes that more difficult. I am in favor of separating information that serves different purposes. Additionally, I am adding hundreds of new lines of examples and additional documentation to many of these source files, which makes the problem much worse.
* Doxygen is certainly more stable the MkDocs and MkDocs is under development. This definitely is a point in favor of Doxygen. However, MkDocs is a small, focused project that is under active development. As such, I'm not overly concerned with a large gap in functionality appearing.
* I disagree that refactoring will be more challenging for a few reasons. 1) The part of the code base that changes often, the plugins, still uses the doxygen documentation scheme. 2) The part of the code base that is relatively stable, the api's, would require editing a markdown file but that could easily be a requirement for merging with master. The issue with invalid links is valid (pun intended :) ). I can't imagine that this has not already been overcome somewhere on this internet though as invalid links have to be a web-wide problem. If it hasn't for some reason the links I use are anchor tags that have to match header id values exactly in the html. It would be fairly trivial to check if a given tag matches any of the available header ids in a html document.
* \cite still needs to exist. The reason it doesn't currently (I believe I just left it as \cite in the docs) is that I don't have any of the cited pdfs yet (should they be distributed with OpenBR?). As soon as I do I can just change \cite to be a link to the proper pdf.
* Search is also really important and MkDocs has an implementation scheduled to land in the next release. Obviously this ties in to MkDocs being less stable. In the interim we could set up google site search, which seems pretty easy to do and gives similar functionality.
* Yes we would lose this functionality. Why is this preferable to distributing static html?
* I want to address this last point in 2 bullets. In this one I will focus on the doxygen's appearance and in the next my general opinions on doxygen. As far as the look goes I think the generated doxygen web page looks dated at best and bad at worst. I also think the problem goes well beyond stylesheets and color choices and is more related to how the information is presented (long scrolling lists with no navigation at all). Also the drop down menus have names that aren't particularly informative. This stuff may sound trivial but if we are serious about attracting people to use and help with OpenBR we need to think about it. I know that if I find docs like ours when I'm trying to use an open source project my interest drops about 200% immediately (and I'm sure I'm not the only one who looks at it that way). In an internet age of material design and bootstrap and all these other beautiful web technologies putting out something that looks like a 1990's government website definitely affects how we are viewed by newcomers who are used to a different level of internet.
* Like I said above I find the doxygen documentation to be poorly designed, difficult to navigate, and often overly confusing. Trying to navigate through pages that have abstract class definitions, hundreds of plugin definitions and a few examples on them is daunting (example: http://openbiometrics.org/doxygen/latest/annotated.html). Obviously, a lot of this can be attributed to poor documentation in OpenBR in general and adding content will alleviate some problems and frustrations. However, I think there is a larger problem with doxygen relating to the content and structure of the docs. As an example @caotto famous class *Idiocy* is currently documented on doxygen and probably shouldn't be. Using doxygen to automatically generate the docs makes it difficult to filter what we document and how it is presented. As OpenBR grows (well over 300 plugins now) presenting information is going to be a larger and larger challenge. Changing now to something that **scales** will save us a lot of pain down the road. In conclusion, I don't think the doxygen appearance problems will be easy to fix (if they are fixable at all) and presenting the documentation like we do, even with additional information, is unacceptable IMO.

Since I have been raging against doxygen I would also like to support MkDocs briefly to highlight that it is a good solution and not just a not-doxygen solution.
* Simple to use. You provide markdown files in a specific directory structure and you have a website.
* Beautiful, modern UI, that is fully hackable.
* Supported by ReadTheDocs (https://docs.readthedocs.org/en/latest/index.html). RTD would allow us to support continuous integration with OpenBR's master branch. Every time a merge is made with master the docs are automatically rebuilt. This solves the problem that OpenBR has of outdated documentation. ReadTheDocs is the reason I chose to use MkDocs in the first place.

So finally to summarize my rant- I think doxygen is ugly and won't scale (from an information presentation perspective) to our needs in the long term. MkDocs and markdown, in general, provide a more customizable, elegant solution which addresses, with RTD integration, the established problem of stale documentation.

caotto:
I guess I'm not completely clear on the scope of the change. I think for e.g. documentation of the parameters of plugins, or saying what a class is for, having comments explaining that in the code is good, but sounds like that behavior will still be retained. 

Personally I never even try to find things by looking at the generated doxygen documents--as Jordan mentions, it's basically a nightmare to navigate. I do think we should consider that some potential documentation problems can be addressed from a policy perspective--we already have a code review policy, so it's just a question of making sure everyone doing reviews is also making sure documentation exists/works correctly.

JordanCheney:
To clarify, the doxygen-style comments will be retained in all plugins. Look at [Adaboost](https://github.com/biometrics/openbr/blob/docs/openbr/plugins/classification/adaboost.cpp) as an example of this. The doxygen comments in places like openbr_plugin.h, openbr.h etc. would be removed in favor of separate markdown files.

jklontz:
@caotto I'm basically interested tallying an up/down vote in favor/against the migration from doxygen to mkdocs. If you and @sklum are in favor (sounds like you are), we'll do the migration, otherwise we won't.

sklum:
Sorry for my delay in responding to this.

@JordanCheney I'm impressed by the changes you've made in this branch (and jealous you have gotten to spend time digging through the entirety of OpenBR).

My thoughts (I do not claim to have gone through every component of this merge so please correct me if I've misunderstood something):
* I support merging the content you've added to the existing documentation, but I think it should go through a review process.  As far as content, I'm most interested in the tutorials and I noticed some typos.  It goes without saying that these are to be expected but having a second eye look over things is important IMO.  Depending on the outcome of this discussion, I'd be willing to do that for this merge.
* Along the lines with the above, I like what you've chosen for a `Getting Started` demo, but there are some minor things I think need to be considered (e.g. Why, when we're using the webcam gallery, do we not need to put the `Open` or `Read` transforms at the beginning of the algorithm string?).
* With regards to searching the documentation, doesn't ReadTheDocs provide this functionality if they host our documentation?
* Not sure about rendering our docs as TeX, but ReadTheDocs seems to support rendering as a PDF.
* I believe we have a quantifiable failure in the OpenBR's current documentation scheme in that we have yet to have a single transform pushed by a non-core developer, and I think _a majority_ of that is due to presentation as opposed to content.  Thus, I support the change to Markdown and mkdocs _if a change to the stylesheet wouldn't sufficiently cover the suggested interface improvements and/or the improvements are much easier to make by migrating to mkdocs_.  Is this the case @JordanCheney?  It sounds like you're making an educated guess at this point.
* @JordanCheney, can you clarify where our doxygen content lives in the new framework?  Is it the `API Docs` and `Plugin Docs` sections of the generated documentation?
* @caotto makes a good point that we should be more diligent in reviews when it comes to documentation.
* Unrelated: It would be cool if we didn't have to write duplicate documentation for `Transform`s that only wrap OpenCV functions.  Is it possible to fetch this documentation automatically and integrate it with ours?

JordanCheney:
@sklum good points.

* A review is 100% necessary. I do my best to keep typos and such out of what I'm adding but I am far from perfect. My original plan for the review process was to draw and break it up and assign it to a few different people so we can have lots of fresh eyes on it. The challenge is definitely going to be in the amount of material.
* Good points on the tutorial. I tried to walk a fine line between interesting and helpful without being too specific. My big lesson learned so far is that writing tutorials is really hard and I would appreciate help in this regard.
* I'm a little bit unclear on the line between MkDocs and ReadTheDocs as far as functionality because I haven't jumped into the integration yet (other then to know it is possible). I believe search comes from the documentation side (MkDocs) however.
* I think the [PanDocs](http://johnmacfarlane.net/pandoc/) project might support markdown->tex directly
* I have read the [doxygen page on website customization](http://www.stack.nl/~dimitri/doxygen/manual/customize.html). My takeaway is that Doxygen supports layout changes as the highest level of in-house customization. For people wanting more than that they recommend using the Doxygen XML output as a base and turning that into HTML.
* The doxygen content from all of the plugin files lives in the <tt>Plugin Docs</tt>. This is possible because the comments are nicely structured and can be automatically converted into markdown at build time. Doxygen comments would continue to be the way plugins are documented. The API docs are populated with doxygen content that has been manually converted to markdown. The doxygen content in the source would then be deleted and the markdown would be the full documentation.
* As far as pulling things directly from OpenCV, I am far from an expert on the ways of the internet. At the very least each OpenCV wrapper should include a reference to the OpenCV function using the <tt>\see</tt> tag. I think RandomForest should have an example of this.

JordanCheney:
I believe I may have found an alternative to MkDocs that would accept doxygen code but can control how the web page is built. I will report on this in the morning!

JordanCheney:
Ok so the alternative I found is called [Breathe](http://breathe.readthedocs.org/) which is recommended on the doxygen customization page. It was designed to take the xml output from doxygen and use it in a documentation engine called Sphinx. Sphinx is the same engine that builds the Python documentation so it is pretty mature and widely utilized. It also is compatible with ReadTheDocs. That is the good news. The bad news is that Breathe requires documentation to be done in a certain way. As an example, **enums** have to be documented inline using 

    /*!< enum option documentation */ 

to be included. Additionally Breath is very rigid in how the output is presented. The CvtTransform looks like this-

cvt

So now we have 2 options. I am still in favor of MkDocs and markdown. The main reason is that I am going to spend hours/days updating all of the documentation (whether we move to MkDocs or not) and I want the output format to be exactly what we decide is most helpful. Locking us into one style of output IMO is basically doing all the work for 1/2 the benefit.

JordanCheney:
Ok @sklum @jklontz @bklare @k-allen @caotto the docs are ready for review round 1. The best way to view them is to pull this branch and then go to ```docs/site``` and run

    python -m SimpleHTTPServer

Then go to http://127.0.0.1:8000 in your browser to browse around. 

2 things to note- Search doesn't work yet but I added the bar in the sidebar so you can see how it will look. I am planning on fixing it this weekend. Second, I have thought about adding 1 more tutorial, Training in OpenBR, and I left a header for it in tutorials but I haven't actually written anything yet, so I would like to hear your thoughts on whether you think it will be helpful to have.

Lastly, I left tags throughout the docs in all caps for things I was unsure about so if you see DOCUMENT ME or FILL ME IN I would appreciate some :)

jklontz:
The compiled html needs to be removed from the PR and hosted elsewhere because:
1. It shouldn't be merged into master
2. It's impossible to review the source code diff currently

Thanks :)

JordanCheney:
It shows the subtraction in the diff now :( Would it be easier for me to just open a new PR?

PR #355, opened 2015-04-10
CropFromMask Tweaks

sklum:
@bklare as the code owner, are you ok with these changes?  I added some edge case checks and refactored a bit because the variable names were confusing (e.g. `w` referred to the number of rows, not cols).


bklare:
@sklum Yeah, mine was a quick and dirty. These look good.

PR #356, opened 2015-04-12
Added border property

sklum:
This branch adds the `border` property which allows a grid to be placed within a region of the entire src matrix as opposed to the whole matrix.  @jklontz any qualms?

### DIFF
--- a/openbr/plugins/metadata/grid.cpp
+++ b/openbr/plugins/metadata/grid.cpp
@@ -37,10 +37,12 @@ class GridTransform : public UntrainableTransform
     void project(const Template &src, Template &dst) const
     {
         QList<QPointF> landmarks;
-        const float row_step = (src.m().rows-(border < 1 ? src.m().rows*border*2 : border*2)) / rows;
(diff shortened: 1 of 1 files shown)


jklontz:
Why is it a float?

sklum:
The border is a percentage of the rows/cols of a matrix.

jklontz:
I suppose I should have figured that out from a closer reading :)

One thing you could do is to also allow `border >= 1` in which case it is in pixels instead of a percentage. That way you can have both uses. LGTM otherwise.

bklare:
I was going to recommend the same thing as @jklontz did.Looks good
otherwise.

sklum:
Good suggestions.  How's this?

jklontz:
Don't mean to be pedantic, but you could factor out `border < 1 ? src.m().rows*border : border` and use it in `row_step` and the for loop.

sklum:
Yes, that should have been done from the get-go.  Should be more pleasing now.

jklontz:
lgtm

PR #357, opened 2015-04-14
Changed the behavior of Factory::parameters() to avoid a segfault

JordanCheney:
In the old version of ```Factory::parameters()``` passing a string with parameter arguments like ```Example(property1=value1)``` caused a segfault. This PR fixes this.

### DIFF
--- a/openbr/openbr_plugin.h
+++ b/openbr/openbr_plugin.h
@@ -946,7 +946,7 @@ struct Factory
     static QString parameters(const QString &name)
     {
         if (!registry) return QString();
-        T *object = make("." + name);
+        QScopedPointer<T> object(make("." + name));
         return object->parameters().join(", ");
     }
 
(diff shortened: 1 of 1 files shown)


openbr/openbr_plugin.h
```diff
@@ -946,8 +946,7 @@ struct Factory
     static QString parameters(const QString &name)
     {
         if (!registry) return QString();
-        QScopedPointer<T> object(registry->value(name)->_make());
-        object->init(name);
+        T *object = make("." + name);
```
jklontz: Use a scoped pointer here to avoid the memory leak.

sklum:
The command `br -gui -algorithm "Open+Grid(6,6,border=.2)+Draw+Show" -enroll ~/openbr/data/family.jpg` works fine for me.

JordanCheney:
This wasn't an issue with the command line it was a problem with how the input to Factory::parameters was handled. If the input wasn't in the registry or had associated metadata the call would segfault

sklum:
Got it.

caotto:
:+1:

jklontz:
lgtm

PR #358, opened 2015-04-14
Updates to trainable distances

sklum:
This branch adds some optimizations in `Distance`s with one or more child `Distance`s such that if none of the child distances are trainable, the distance returns `false` for its `trainable` method skipping the need to project enrollment during training.  There is also some related cleanup of `FuseDistance`.

### DIFF
--- a/openbr/plugins/distance/sum.cpp
+++ b/openbr/plugins/distance/sum.cpp
@@ -26,19 +26,9 @@ namespace br
  * \brief Sum match scores across multiple distances
  * \author Scott Klum \cite sklum
  */
-class SumDistance : public Distance
+class SumDistance : public ListDistance
 {
     Q_OBJECT
-    Q_PROPERTY(QList<br::Distance*> distances READ get_distances WRITE set_distances RESET reset_distances)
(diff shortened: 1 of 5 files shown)


openbr/plugins/distance/sum.cpp
```diff
@@ -32,6 +32,14 @@ class SumDistance : public UntrainableDistance
     Q_PROPERTY(QList<br::Distance*> distances READ get_distances WRITE set_distances RESET reset_distances)
     BR_PROPERTY(QList<br::Distance*>, distances, QList<br::Distance*>())
 
+    bool trainable()
```
jklontz: Should SumDistance not inherit from UntrainableDistance?

jklontz: Also, seems like there's some common logic for distances with a `Q_PROPERTY(QList<br::Distance*> distances READ get_distances WRITE set_distances RESET reset_distances)` as the first member, where they all implement trainable() in the same way? Worth factoring that out into a new base class?

sklum:
Much cleaner.  Any other thoughts @jklontz?

jklontz:
lgtm

PR #359, opened 2015-04-17
CrossValidation Refactor

sklum:
This branch introduces changes to the `crossValidate` framework such that it works with changes that have been made to `br -train`,  `br -enroll` and  `br -compare`.

Relevant changes:
* Dropped support for leaveOneOut cross-validation.  The code was convoluted and (I'm assuming) used only by me.
* Moved partitioning logic from `TemplateList::FromGallery` to `TemplateList::partition`.
* Partitioning now occurs _within_ `CrossValidateTransform` via `TemplateList::partition`, but `TemplateList::partition` is still called from within `TemplateList::FromGallery` (such that things like `.eval` output with crossValidation still work).
* Added better descriptions for the `allPartition` and `duplicatePartition` flags that can be used with `crossValidate`.

As a reminder, users can leverage `crossValidate` to create a model:
`br -crossValidate 2 -algorithm "CrossValidate(your_trainable_algorithm):CrossValidate+Dist(L2)" -train your_query_data.xml your_target_data.xml your_model`

To enroll using that model:
`br -crossValidate 2 -algorithm your_model -enroll your_query_data.xml your_query_data.gal`
`br -crossValidate 2 -algorithm your_model -enroll your_target_data.xml your_target_data.gal`

To compare the enrolled galleries using `evalOutput` (note that this command will generate two csvs corresponding to each partition):
`br -crossValidate 2 -algorithm your_model -compare your_target_data.gal your_query_data.gal your_results%1.eval[target=your_target_data.gal,query=your_query_data.gal]`

The splits generated for cross-validation are currently based on the `Label` field (although the groundwork is there to deviate from this restriction in the future) and are consistent across multiple executions of `br`.


caotto:
What I would like to see, is for us to automatically attack CrossValidateTransform/Distance when training is invoked, with -crossValidate N supplied. Feels very redundant to specify -crossValidate N, and also add CrossValidate wrapper transform, and also add CrossValidate distance metric

jklontz:
Agreed w/ @caotto, though I don't see this as a blocker for merging the current work.

sklum:
I like the idea as well @caotto.  There are some simple tweaks I'll need to make.  My only concern is that, in the experiments I've been doing, putting `CrossValidate` as the root as opposed to after `Open` (and `Scale` in my case).  I'll investigate this more.

jklontz:
I was thinking that it could be intelligently inserted around only the transforms that actually require training?

sklum:
I think there is some issue with it not being root or near root? @caotto?

caotto:
If the algorithm string is like "A+B+C", I feel like it should be possible to parse the pipe, and insert prior to the first non-trainable transform.

@sklum: I do think that you and I discussed doing this previously, but you ran into some kind of problem, but I don't remember the details at all.

sklum:
Going to merge this for now.  Will address the above suggestions when I have some more free time.

PR #360, opened 2015-04-22
Add IET, changed ROC sampling

bhklein:
This PR introduces the Identification Error Tradeoff (IET) curve as described in the [NIST FRVT 2013] (http://biometrics.nist.gov/cs_links/face/frvt/frvt2013/NIST_8009.pdf) report for evaluating the open-set identification problem.  

Additionally, FAR/FPIR values are now sampled in a range based on the order of magnitude of impostor comparisons/searches.  This was done in order to keep consistency of sampling points across testing splits, which are assumed to generally have the same order of magnitude of comparisons.

### DIFF
--- a/openbr/core/eval.cpp
+++ b/openbr/core/eval.cpp
@@ -196,7 +196,10 @@ float Evaluate(const Mat &simmat, const Mat &mask, const QString &csv, const QSt
     QList<Comparison> comparisons; comparisons.reserve(simmat.rows*simmat.cols);
 
     // Flags rows as being mated or non-mated searches
+    // Positive value: mated search, negative value: non-mated search
+    // Value of 0: ignored search
     QVector<int> genuineSearches(simmat.rows, 0);
+
     int totalGenuineSearches = 0, totalImpostorSearches = 0;
     int genuineCount = 0, impostorCount = 0, numNaNs = 0;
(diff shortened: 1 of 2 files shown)


jklontz:
What page number is IET defined on? Can't seem to find it...

openbr/core/eval.cpp
```diff
@@ -193,6 +194,10 @@ float Evaluate(const Mat &simmat, const Mat &mask, const QString &csv, const QSt
 
     // Make comparisons
     QList<Comparison> comparisons; comparisons.reserve(simmat.rows*simmat.cols);
+
+    // Flags rows as being mated or non-mated searches
+    QVector<int> genuineSearches(simmat.rows, 0);
```
jklontz: I think the comment should be updated to explain why this isn't a boolean vector.

jklontz:
lgtm otherwise, thanks!

bhklein:
FPIR and FNIR are defined starting on page 17.  I used IET instead of DET to discriminate from the DET for FAR vs FRR we already compute in -eval.  IET is used at least once in the report, namely in figure 6 on page 25.

jklontz:
ok cool, thanks

bhklein:
Thanks, I'll merge at the end of the day if no one else has anything to add.

PR #361, opened 2015-04-23
Blend tweak

sklum:
Added support for more than two matrices in blend.  @jklontz, tell me what I'm missing :smile:.

### DIFF
--- a/openbr/plugins/imgproc/blend.cpp
+++ b/openbr/plugins/imgproc/blend.cpp
@@ -23,7 +23,7 @@ namespace br
 
 /*!
  * \ingroup transforms
- * \brief Alpha-blend two matrices
+ * \brief Alpha-blend matrices
  * \author Josh Klontz \cite jklontz
  */
 class BlendTransform : public UntrainableMetaTransform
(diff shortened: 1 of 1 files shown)


openbr/plugins/imgproc/blend.cpp
```diff
@@ -29,13 +29,22 @@ namespace br
...
-        addWeighted(src[0], alpha, src[1], 1-alpha, 0, dst);
+        dst.file = src.file;
+
+        if (src.size() != 2) {
+            dst.m() = Mat::zeros(src.m().rows, src.m().cols, src.m().type());
+            for (int i=0; i<src.size(); i++) {
+                const float weight = (alpha.size() == src.size()) ? alpha[i] : 1./src.size();
+                dst.m() += src[i]*weight;
+            }
+        } else {
```
jklontz: I feel like with a change in the semantics of alpha, this could be re-written such that the `else` block is removed, and 2 is just a special case of N?

jklontz:
You missed updating the doxygen docs :p

sklum:
Ty.

PR #363, opened 2015-04-27
OpenCV Cascade Integration

JordanCheney:
@bklare @sklum @jklontz I'm opening this as a checklist for cascade integration so you all can stay up to date on the progress-

### Backend
- [x] Integrate training code using the filesystem as an intermediary (i.e write to disk with OpenBR and then read with the OpenCV train code)
- [x] Remove the file system middle man so images can be preprocessed in OpenBR and then passed directly to train code
- [x] Incorporate representations instead of OpenCV features into the backend
- [x] Fit the backend code into the Classification / SlidingWindow paradigm
- [x] Convert to OpenBR I/O

### Frontend
- [x] Implement the sliding window front end
- [x] Incorporate representations instead OpenCV features
- [x] Add to sliding window / classification
- [x] Convert to OpenBR I/O


jklontz:
:+1:

sklum:
Sounds good @JordanCheney - looking forward to the progression.

JordanCheney:
Updated the checklist to reflect an easier path

sklum:
@JordanCheney IMO, this can and should be merged ASAP.

openbr/plugins/representation/square.cpp
```diff
@@ -0,0 +1 @@
+
```
jklontz: extraneous file?

openbr/core/boost.h
```diff
@@ -0,0 +1,112 @@
...
+    float minHitRate;
+    float maxFalseAlarm;
+
+    CascadeBoostParams();
+    CascadeBoostParams(int _boostType, float _minHitRate, float _maxFalseAlarm,
+                       double _weightTrimRate, int _maxDepth, int _maxWeakCount);
+    virtual ~CascadeBoostParams() {}
+};
+
+struct CascadeBoostTrainData : CvDTreeTrainData
```
jklontz: Seems to me that `CascadeBoostTrainData` and `CascadeBoostTree` need not be exposed in this header file as they are only used in `boost.cpp`.

openbr/plugins/representation/square.cpp
JordanCheney: Yep, removed.

openbr/core/boost.h
JordanCheney: ```CascadeBoostTree``` is used by ```BoostedForest```. ```CascadeBoostTrainData``` is moved to ```boost.cpp```

PR #364, opened 2015-04-27
FTE in File::append

sklum:
In an algorithm string along the lines of `Transform1+Transform2/Transform3+Transform4`, if either `Transform2` or `Transform3` sets the `fte` flag, it will not be present in the `Template`s passed to `Transform4` because of the logic in the various `Fork::project` calls.  This is because the `Template`s passed through a given child `Transform` of fork are merged with the original file and nothing handles the `fte` flag in these functions.  This branch changes that.

### DIFF
--- a/openbr/openbr_plugin.cpp
+++ b/openbr/openbr_plugin.cpp
@@ -97,6 +97,7 @@ void File::append(const File &other)
         }
     }
     append(other.m_metadata);
+    fte = fte | other.fte;
 }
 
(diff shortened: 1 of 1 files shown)


sklum:
@jklontz @bklare @caotto let me know if you have any thoughts.

caotto:
:+1:

jklontz:
:+1:

bklare:
Great improvement.

PR #365, opened 2015-04-27
Distance training with FTEs

sklum:
FTE templates shouldn't be used to train distances.

### DIFF
--- a/openbr/core/core.cpp
+++ b/openbr/core/core.cpp
@@ -86,12 +86,15 @@ struct AlgorithmCore
             qDebug("Projecting Enrollment");
             trainingWrapper->projectUpdate(data,data);
 
-            for (int i=data.size()-1; i>=0; i--)
-                if (data[i].file.fte || data[i].file.getBool("FTE"))
-                    data.removeAt(i);
+            TemplateList distanceData;
+            for (int i=0; i<data.size(); i++)
+                if (!data[i].file.fte && !data[i].file.getBool("FTE"))
+                    distanceData.append(data[i]);
+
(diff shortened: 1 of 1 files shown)


openbr/core/core.cpp
```diff
@@ -86,6 +86,10 @@ struct AlgorithmCore
             qDebug("Projecting Enrollment");
             trainingWrapper->projectUpdate(data,data);
 
+            for (int i=data.size()-1; i>=0; i--)
+                if (data[i].file.fte || data[i].file.getBool("FTE"))
+                    data.removeAt(i);
```
caotto: Doing deletes of interior elements like this in a flat structure like a QList is typically not efficient (better to create a new list to contain non-deleted items, copy them out to the new list, drop the old list), although the perf impact may not be noticeable in this case.

jklontz: Do you have a source for that? A QList isn't a vector, so I would expect removeAt to be amortized O(1).

caotto: For example this article explicitly talks about QList

https://freininghaus.wordpress.com/2013/07/18/some-thoughts-on-algorithmic-complexity-part-2-be-careful-if-you-insert-into-or-remove-from-a-stdvector-qvector-or-qlist-in-a-loop/

How do you amortize deletes at arbitrary points, while maintaining integer
indexing?

jklontz: Thanks for the link

sklum:
@caotto Better?

caotto:
yes

PR #366, opened 2015-04-27
Not storing cv::Mat metadata when fte

sklum:
@caotto @jklontz per our earlier convo.  My expectation is that the algorithm string should be able to handle removal of specific metadata in the case of successful enrollment.  This change constitutes a catch-all in the case of an `fte`.

### DIFF
--- a/openbr/plugins/gallery/binary.cpp
+++ b/openbr/plugins/gallery/binary.cpp
@@ -178,10 +178,13 @@ class galGallery : public BinaryGallery
         else if (t.file.fte) {
              // Only write metadata for failure to enroll, but remove any stored QVariants of type cv::Mat
             File f = t.file;
-            QList<QVariant> values = f.localMetadata().values();
-            for (int i=0; i<values.size(); i++)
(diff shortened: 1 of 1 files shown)


openbr/plugins/gallery/binary.cpp
```diff
@@ -175,8 +175,15 @@ class galGallery : public BinaryGallery
     {
         if (t.isEmpty() && t.file.isNull())
             return;
-        else if (t.file.fte)
-            stream << Template(t.file); // only write metadata for failure to enroll
+        else if (t.file.fte) {
+             // Only write metadata for failure to enroll, but remove any stored QVariants of type cv::Mat
+            File f = t.file;
+            QList<QVariant> values = f.localMetadata().values();
+            for (int i=0; i<values.size(); i++)
```
jklontz: This is arguably a good case for a [QMapIterator](http://doc.qt.io/qt-5/qmapiterator.html) (or STL iterator) to avoid the value-to-key lookup.

sklum: Good call.

PR #367, opened 2015-04-29
Eval/plotting fixes

bhklein:
This eliminates the use of geom_smooth in the PRE/REC plot, and allows the user to specify a minimum FAR/FRR.  Also fixes table display in the case where file pivots are used without smoothing.

### DIFF
--- a/openbr/core/eval.cpp
+++ b/openbr/core/eval.cpp
@@ -232,7 +232,6 @@ float Evaluate(const Mat &simmat, const Mat &mask, const File &csv, const QStrin
     if (numNaNs > 0) qWarning("Encountered %d NaN scores!", numNaNs);
     if (genuineCount == 0) qFatal("No genuine scores!");
     if (impostorCount == 0) qFatal("No impostor scores!");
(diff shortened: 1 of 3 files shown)


openbr/core/eval.cpp
```diff
@@ -232,7 +232,7 @@ float Evaluate(const Mat &simmat, const Mat &mask, const QString &csv, const QSt
     if (numNaNs > 0) qWarning("Encountered %d NaN scores!", numNaNs);
     if (genuineCount == 0) qFatal("No genuine scores!");
     if (impostorCount == 0) qFatal("No impostor scores!");
-    if (totalImpostorSearches == 0) totalImpostorSearches = 1;
+    if (totalImpostorSearches == 0) totalImpostorSearches = 10;
```
jklontz: What's the deal here?

jklontz:
LGTM otherwise

openbr/core/eval.cpp
bhklein: Was meant to avoid the case where expFPIR=0, but I think this solution works better

PR #369, opened 2015-05-12
Random test/train split in TemplateList::Partition

bhklein:
This PR allows the user to specify a `bootStrap` seed to randomly split a dataset into training/testing sets based on a partition `inputVariable`.  

This could be used as a quick one-off with a seed of your choosing, or possibly over a set of numbers, say (1..10) within a script, for example:
```
for i in {1..10}
do
  br -crossValidate 1 -algorithm "CrossValidate(description=$ALG,bootStrap=${i}):CrossValidate+Dist" -train target.xml query.xml model
  br -crossValidate 1 -algorithm model -enroll target.xml target.gal
  br -crossValidate 1 -algorithm model -enroll query.xml query.gal
  br -crossValidate 1 -algorithm model -compare target.gal query.gal results%1.eval
  mv results0.csv Algorithm_Split/results_${i}.csv
done
br -plot Algorithm_Split/* results.pdf[smooth=Split]
```
This is useful for datasets where the testing size becomes prohibitively small as n increases in n-fold cross-validation.  

@sklum please review, @bklare fyi

### DIFF
--- a/openbr/openbr_plugin.cpp
+++ b/openbr/openbr_plugin.cpp
@@ -529,20 +529,22 @@ QList<int> TemplateList::applyIndex(const QString &propName, const QHash<QString
     return result;
 }
 
-TemplateList TemplateList::partition(const QString &inputVariable, unsigned int bootStrap) const
+TemplateList TemplateList::partition(const QString &inputVariable, unsigned int randomSeed) const
 {
     TemplateList partitioned = *this;
-    if (bootStrap > 0) {
-        // Use separate RNG from Common::randN() to avoid re-seeding the global RNG.
-        // rng is seeded with the inputVariable hash in order to maintain partition across br runs, given the same bootStrap seed.
-        RandomLib::Random rng;
+    const int crossValidate = Globals->crossValidate;
+
+    // Use separate RNG from Common::randN() to avoid re-seeding the global RNG.
+    // rng is seeded with the inputVariable hash in order to maintain partition across br runs, given the same randomSeed seed.
+    RandomLib::Random rng;
+    if (crossValidate == 1) {
         for (int i=0; i < partitioned.size(); i++) {
             const QByteArray md5 = QCryptographicHash::hash(partitioned[i].file.get<QString>(inputVariable).toLatin1(), QCryptographicHash::Md5);
-            quint64 seed = md5.toHex().right(8).toULongLong(0, 16) + bootStrap;
-            rng.Reseed(seed);
-            float result = rng.FloatN();
+            quint64 labelSeed = md5.toHex().right(8).toULongLong(0, 16) + randomSeed;
+            rng.Reseed(labelSeed);
 
-            // Roughly 2/3rd to training, 1/3rd to testing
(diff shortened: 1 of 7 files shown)


openbr/openbr_plugin.cpp
```diff
@@ -528,28 +529,47 @@ QList<int> TemplateList::applyIndex(const QString &propName, const QHash<QString
...
+        // rng is seeded with the inputVariable hash in order to maintain partition across br runs, given the same bootStrap seed.
+        RandomLib::Random rng;
+        for (int i=0; i < partitioned.size(); i++) {
+            const QByteArray md5 = QCryptographicHash::hash(partitioned[i].file.get<QString>(inputVariable).toLatin1(), QCryptographicHash::Md5);
+            quint64 seed = md5.toHex().right(8).toULongLong(0, 16) + bootStrap;
+            rng.Reseed(seed);
+            float result = rng.FloatN();
+
+            // Roughly 2/3rd to training, 1/3rd to testing
+            if (result <= (2.0/3.0))
```
sklum: I'd like to maintain the ability to do `n` fold cross-validation with the bootstrap option.  So, generate a partition between 0 and `crossValidate` as is done below, except in the `bootStrap` case, base that partition on the random generator (and seeding system) you already have set up.

bklare:
Haven't reviewed the code but really excited about this PR. Thank you!

openbr/openbr_plugin.cpp
bhklein: @sklum How's this?  I've changed it to do the 1/3rd 2/3rd split for the -crossValidate 1 case.  Otherwise it returns a random partitioning if the seed is >0

sklum:
@bhklein I'm not sure why I can't see your comment on here.

From an evaluation perspective, the question is: is there a use case where you would want to do k-fold crossvalidation but only use a single set of splits for evaluation.  I would argue that there isn’t a good reason for this, and we still support that option without adding this extra logic (the user can simply use only the results from the split they are interested in).

Having said this, from training time standpoint, I could see wanting to do crossvalidation but only having the time to train a model based on one of the splits.  In this case, though, we should still support the option for k-fold crossvalidation (i.e. other partitioning schemes besides 2/3 training 1/3 testing).  

My suggestion for this is to allow for an argument to crossValidate to be negative.  If negative, this indicates that that we want to generate k-fold cross validate splits but only test on the 0th fold (and only train once on the remaining folds).  Thus, you can remove the special `crossValidate == 1` logic within `TemplateList::partition`.  Then, you’d need to add the logic within that function and tweak some of the stuff you have changed in `CrossValidate::train` and `CrossValidate::project`.

After you make these changes, I’ll fix some of the logic in other parts of the codebase to handle the negatives.  Sound good?

bhklein:
@sklum here's round three.  `-` is parsed as a parameter instead of a function from the command line if it can be converted to an `int`.

sklum:
Looks great @bhklein. Nice work. Go ahead and test this as we discussed.  Also, never write a commit message while having a conversation like I just did.

PR #371, opened 2015-05-29
Support file in memory buffer

clcarwin:
We can query and transform files in memory buffer by this patch:

QString path("memory:xxx123xxx");
std::vector<char> v;
v.insert(v.end(),buffer,buffer+bufferlen);
br::File::MemoryFile.insert(path,v);

br::Template query(path);
br::TemplateList queries;
queries.append(query);

queries >> *transform;
...
br::File::MemoryFile.remove(path);


jklontz:
Thanks for the PR, but I don't see how this is a useful feature. You can already append to the Template directly: `query.append(imdecode(buffer));` which is identical to what you accomplish with this extra level of indirection.

clcarwin:
Thanks. I have got a deeper understanding of the Template and OpenTransform class.

PR #374, opened 2015-06-05
Documentation for cascade branch transforms

JordanCheney:
This should have been included in the merge of the cascade branch. @sklum you are probably the best reviewer it should be quick.


sklum:
Looks good @JordanCheney.

PR #375, opened 2015-06-05
fix to dlib cmake file

JordanCheney:
Dlib cmake file now conforms to the new cmake structure. @sklum tapping you for review


sklum:
:+1:

PR #378, opened 2015-06-08
automatically build dlib by default

jklontz:
@sklum please review

@JordanCheney heads up


sklum:
lgtm

JordanCheney:
If dlib is being built by default should it be included in 3rdparty? If not should we add install instructions to the docs for getting it?

JordanCheney:
Nevermind didn't see the download part of the cmake script. Very cool. This is what I get for reviewing pull requests on my phone :)

PR #382, opened 2015-06-10
removed frvt2012 API

jklontz:
@JordanCheney I think you're the only one that might remotely care. This ok with you?


jklontz:
@JordanCheney poke

JordanCheney:
Looks good!

PR #383, opened 2015-06-15
Filter metadata

bklare:
A simple transform for filtering based on metadata.

### DIFF
--- a/openbr/plugins/metadata/filtermetadata.cpp
+++ b/openbr/plugins/metadata/filtermetadata.cpp
@@ -23,9 +23,9 @@ namespace br
  * \ingroup transforms
  * \brief Filters templates such that the only remaining template wills have metadata values witihin the
  *          the specified ranges. 
- * \br_property key1,key2 The meta-data key(s) to filter on
- * \br_property value1,value2 The values to compare the values of key1 and key2 entires against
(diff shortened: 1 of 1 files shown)


openbr/plugins/metadata/filtermetadata.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <openbr/plugins/openbr_internal.h>
+
+namespace br
+{
+
+/*!
+ * \ingroup transforms
+ * \brief Filters templates such that the only remaining template wills have metadata values witihin the
+ *          the specified ranges. 
+ * @param[in] key1,key2 The meta-data key(s) to filter on
```
jklontz: @JordanCheney does the doc engine care about `\param` vs. `@param`?

JordanCheney: Right now the doc engine parses only ```\br_property``` which I am using as a replacement for param.

jklontz:
@bklare LGTM pending doc fix

bklare:
Thanks for the clarification. Just updated.

JordanCheney:
@bklare one more change ```\br_property```expects input as ```\br_property type name description``` so in your case ```\br_property QString key1 description```. Sorry for the confusion.

bklare:
@JordanCheney Thanks for the clarification.

@jklontz Should be GTG.

PR #384, opened 2015-06-15
Serialize algorithm

bklare:
Allows the serialization of OpenCV objects that derive from the Algorithm object. Currently only the CvStatModel is supported.


jklontz:
lgtm

PR #385, opened 2015-06-15
Output to gallery that can be ingested into matlab

bklare:
This is a new gallery function for quick and dirty export of templates into a format easily ingestable in other environments (e.g., matlab).


bklare:
This is being merged.

PR #387, opened 2015-06-24
Online Rank-Order distance clustering

JordanCheney:
Transform to perform Rank-Order distance clustering.

### DIFF
--- a/openbr/plugins/cluster/onlinerod.cpp
+++ b/openbr/plugins/cluster/onlinerod.cpp
@@ -15,6 +15,7 @@ namespace br
  * \br_property int kNN Maximum number of nearest neighbors to keep for each template. Default is 20.
  * \br_property float aggression Clustering aggresiveness. A higher value will result in larger clusters. Default is 10.
  * \br_property bool incremental If true, compute the clusters as each template is processed otherwise compute the templates at the end. Default is false.
+ * \br_property QString evalOutput Path to store cluster informtation. Optional. Default is an empty string.
  * \br_paper Zhu et al.
  *           "A Rank-Order Distance based Clustering Algorithm for Face Tagging"
  *           CVPR 2011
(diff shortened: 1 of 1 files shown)


JordanCheney:
@caotto I think you are the best reviewer here

openbr/plugins/cluster/onlinerod.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    void identifyClusters(TemplateList &dst)
+    {
+        Clusters clusters = ClusterGraph(neighborhood, aggression, evalOutput);
+        if (Globals->verbose)
+            qDebug("Built %d clusters from %d templates", clusters.size(), templates.size());
+
+        for (int i = 0; i < clusters.size(); i++) {
+            // Calculate the center of each cluster
```
caotto: Different measures of center are possible, you are specifically using the centroid here, so you should say that

openbr/plugins/cluster/onlinerod.cpp
```diff
@@ -0,0 +1,124 @@
...
+                center += converted;
+            }
+            center /= clusters[i].size();
+
+            // Calculate the Euclidean distance from the center to use as the cluster confidence
+            foreach (int t, clusters[i]) {
+                templates[t].file.set("Cluster", i);
+                Mat c; templates[t].m().convertTo(c, CV_32F);
+                Mat p; pow(c - center, 2, p);
+                templates[t].file.set("ClusterConfidence", sqrt(sum(p)[0]));
```
caotto: since sqrt is order preserving, it's typically unnecessary to actually calculate it when dealing with Euclidean distance.
If you want to look at the values manually maybe it's nicer to have smaller numbers? if you're just going to use this in some automatic application, you can omit sqrt

caotto:
Couple minor comments, otherwise seems fine

openbr/plugins/cluster/onlinerod.cpp
JordanCheney: Fixed thanks

openbr/plugins/cluster/onlinerod.cpp
JordanCheney: Good to know. I think I will leave it for now because the confidence will be used in some manual review type applications

JordanCheney:
All fixed. Thanks!

PR #388, opened 2015-06-24
Remove maxCatCount from Classifier

JordanCheney:
Remove unnecessary function maxCatCount() from classifiers. @sklum can you review?


sklum:
Looks good to me.  Was gunna ask the point of this function but forgot to before I left :smile:.

PR #389, opened 2015-06-26
Caffe

JordanCheney:
This is a first pass and shallow integration of Caffe into OpenBR. It requires Caffe to already be installed. Currently both train and project require data be stored on disk and loaded within Caffe. The next step for integration IMO is to be able to pass information into the Net at project time, I left a comment where I think that should occur. @jklontz you previously expressed interest in reviewing this so I'm tapping you :). @imaus10 FYI.


caotto:
A few comments:

I'm not sure that running training through openbr vs. the caffe application is of any value (their checkpointing is very good, and you will probably want to use it to stop/start training to adjusting the learning rate during training). 

I'm under the impression that doing extraction one image at a time with Caffe is grossly inefficient, and that ideally it should be done in small batches (say neighborhood of 128 images at a time). 

As far as passing data to Caffe, as far as I'm aware the preferred method for doing that is to define your input layer as a MemoryDataLayer (type: "MemoryData" in the net definition), and then access that layer through the C++ API 

I think re-instantiating the Net for every project call is probably also inefficient, and that that would be better done in an initialize method.

caotto:
As far as interoperability with caffe goes, I also have a gallery for reading/writing lmdb databases (in the style supported by caffe), although it is perhaps not up to current coding/documentation standards. 

I could upload that, if you are interested in cleaning up/merging it.

JordanCheney:
Pretty much agree on all points, I think training in OpenBR would be useful if some preprocessing is required that isn't supported by Caffe, perhaps the MemoryData layer would work there as well, although with large training sets that don't fit in memory this could present challenges. Batch processing should be implemented for efficiency, I would call this implementation naive at best :). A lmdb gallery would be awesome I was about to start writing one so you would be saving me some time.

caotto:
Well, basically you can preprocess your training set in openbr, then write it to an lmdb database, which you can then load from caffe (and if you are doing SGD, you will load from the gallery in batches anyway).

I'll upload my lmdbGallery implementation a little later tonight

JordanCheney:
I think reimplementing as a TimeVaryingTransform with the net initialized on init, no training, and Forward only called when a certain number of templates have been accumulated would solve these problems no?

caotto:
More or less.

That approach does open you up to an issue with FTE handling. Basically if you have something like:
Preprocessing+Caffe+OtherThings
and an FTE occurs during the preprocessing steps, file order will not be maintained if you batch the templates like that in the CaffeTransform (since FTE templates are removed from normal execution, they will not be batched by CaffeTransform, so will end up out of order w/r/t the templates that are batched). 

So this matters if you are doing evaluation, and expect a fixed order to be maintained after enroll. 

So you can either fix the order afterwards, or handle batching differently. This could be done by handling batching at the stream level, rather than internally to CaffeTransform, so batching is done in terms of Stream's FrameData structures, rather than just templates. I haven't gotten around to doing anything about this, and don't really have the availability at this moment.

caotto:
k I pushed the lmdb plugin. It is pretty rough! Bear in mind it was written using badly configured emacs.

jklontz:
@JordanCheney No comments at the moment, sounds like are on the right track :) Glad to see this being worked on!

bklare:
Excellent, Jordan. Glad to see this integration.

JordanCheney:
@caotto @jklontz I'm running into an issue instantiating the Caffe::Net at plugin initialization time. The copy and assignment operators for the Net are set to private. This is implemented [here](https://github.com/BVLC/caffe/blob/master/include/caffe/net.hpp#L262) and [here](https://github.com/BVLC/caffe/blob/master/include/caffe/common.hpp#L31). Have either of you seen this before and are aware of workaround?

caotto:
Is there a problem with dynamically allocating the Net in initialize so
that you don't have to copy/assign it?

JordanCheney:
No that took care of it. Thanks!

JordanCheney:
@caotto @jklontz I think this is ready for merge can either of you review?

jklontz:
Been out of the loop on this one so I'll differ to @caotto if he wants to review. From the cursory glance I made, it LGTM.

caotto:
What is the expected use, in terms of concurrency? 

Is it safe to make concurrent calls to the same net like this? I don't know offhand but this syntax:
+        data_layer->AddMatVector(src.toVector().toStdVector(), std::vector<int>(src.size(), 0));
inspires doubt

caotto:
Also, for memory data layer, don't you have to specify a fixed batch size in the network definition? Are you expected to honor that when using it like this?

JordanCheney:
This is definitely not thread safe. I'll add an explicit check that parallelism = 0 before executing. A batch size does have to be specified in the network definition and it needs to be matched by the number of mats in src. This is guaranteed by the check at the beginning of the project function.

caotto:
Did you check the performance impact of doing single image vs. batch? I definitely recall reading that it was inefficient to do one image at a time, but didn't do any profiling

caotto:
Also, I don't agree with requiring parallelism = 0, because that also limits use of multiple cores to do preprocessing (while presumably running caffe on GPU). You can just make this timeVarying (so that it will be called single-threaded), you don't have to do explicit batching if you don't want to deal with the FTE issue.

JordanCheney:
I haven't done any official profiling but I've been running it with a single image and have encountered sufficiently fast runtimes thus far. I'm not exactly sure what you mean by explicit batching, I think the current set-up allows for batch processing it just falls on the user to format the input template and properly interpret the output. Finally, TimeVarying sounds like the proper way to handle the threading issue so I'll make that change.

caotto:
I don't know what sufficiently fast means, I'm basically wondering what the
performance difference between N 1 template enrollments, vs. 1 N template
batch enrollment is.

With regards to batching, I was referring the previously discussed idea of
grouping templates in this transform. If you don't want to deal with it,
that's fine.

JordanCheney:
Its been on the order of milliseconds per image to run. As soon as I get the chance I'll so some profiling to check on performance changes

caotto:
What hardware?

JordanCheney:
GeForce GT 750M GPU (the GPU in my laptop)

caotto:
Have you looked at CPU mode performance?

caotto:
Anyway, if non-batched extraction is not a major performance concern, you
could also try making nThreads copies of the net rather than using
single-threaded processing of a single net

JordanCheney:
I haven't looked at CPU performance but its on my to-do list. I don't think multiple copies of the net is a good solution for the GPU since we are normally memory constrained but for the CPU it could be a good solution. Could this be down using an OpenBR resource?

caotto:
Yes, that is what resource is for.

It's possible to switch modes (timeVarying or not) by reimplementing
timeVarying(), so you could default to non-timeVarying for CPU mode, and
timeVarying for GPU mode (and also allow explicit override if someone has a
big GPU or whatever).

JordanCheney:
@caotto How does it look now?

caotto:
Looks OK, feel free to merge if you have done appropriate testing (does default projectUpdate work correctly etc.)

JordanCheney:
Tested on CPU and GPU with good results so I'm going to go ahead and merge

bklare:
@JordanCheney Nice job on this integration of Caffe. I was trying this out and ran into an issue. Perhaps you can help.

The issue revolves around how this transform seems to operate. From my understanding, based on the parameter `batch_size`, the transform expects the template to contain a list of `batch_size` parameters. For example, I pulled a pre-trained model from the zoo and this value was 1,728. I created a little dummy transform that uses a call to `project(TemplateList ...)` to poll together batch_size templates into a single template, and then calls CaffeFV.project(). However, when I make this call it being routed through InpendentTransform, which pulls out all of the matrices from the list, puts them in single matrix templates, and calls CaffeFV. In turn, the qFatal on line 96 is hit.

Any feedback you can provide here would be appreciated. I am not sure if there is some other approach you are using to run this transform.

caotto:
Oh, well hitting independent can be avoided simply by setting the
appropriate variable in the constructor, or inheriting
UntrainableMetaTransform.

bklare:
Charles - thank you for the quick reply. The little test class I created is
already deriving from UntrainableMetaTransform.

Here is a the code:

#include <openbr/plugins/openbr_internal.h>

namespace br
{
    class CaffeTestTransform : public UntrainableMetaTransform
    {
        Q_OBJECT

        Transform * caffe;

        void init()
        {
            caffe =
Transform::make("CaffeFV(VGG_ILSVRC_16_layers_deploy.prototxt,VGG_ILSVRC_16_layers.caffemodel)");
        }

        void project(const Template &src, Template &dst) const
        {
            qWarning("Single template project not supported by
VerifyDetectionTransform.");
            dst.file.fte = true;
        }

        void project(const TemplateList &src, TemplateList &dst) const
        {
          //  QScopedPointer<Transform> caffe(Transform::make("CaffeFV"));
            int nLayers = 1728;

            if (src.size() == 0) {
                qDebug() << "No templates";
                return;
            }

            Template srcCaffe(src[0].m());
            for (int i = 1; i < nLayers; i++)
                srcCaffe.append(src[0].m().clone());

            int last = 0;
            Template dstCaffe;
            for (int i = 0; i < src.size(); i++) {
                int idx = i % nLayers;

                srcCaffe.replace(idx, src[i]);


                if ((i - 1) % nLayers == 0)
                {
                    caffe->project(srcCaffe, dstCaffe);
                    for (int j = last; j <= i; j++)
                        dst.append(dstCaffe[j]);
                    last = i;
                }
            }

            if (last != src.size())
            {
                caffe->project(srcCaffe, dstCaffe);

                for (int i = last; i <= src.size(); i++)
                    dst.append(dstCaffe[i]);
            }
        }
    };

    BR_REGISTER(Transform, CaffeTestTransform)

} // namespace br

caotto:
No, the caffe transform should inherit UntrainableMetaTransform

bklare:
I see. Thanks for the clarification.

JordanCheney:
Did this fix the problem for you @bklare? I can push a fix, this is a good catch

bklare:
Thanks for asking. It did. I would have a pushed a fix, but I am having
some other issues, seeemingly due to the dimensions of the input image. I
am abandoning this test for now as it does not appear Caffe is well suited
for a scalable openbr integration at this point.

JordanCheney:
Hmmmm, I'd like to follow up with you on the errors you're seeing so that I can update the transform. Can you send me the link to the model you downloaded from the model zoo? If you are using the exact model you downloaded, the issue is probably in the data layer. The OpenBR integration requires a memory data layer to take images directly from memory and not from disk. Can you give me any more information about the issue you're seeing?

caotto:
@bklare: why do you think so?

bklare:
The error I hit is on line 100 of the CaffeFV file. It seems to be related to the input file.  Below is the model architecture. It seems to call for an image that is 224x224. I kept it RGB due b/c there is another input dimension of 3. However, in looking now, there is a fourth input dimension of 10 that may be causing the issue. 

https://gist.githubusercontent.com/ksimonyan/211839e770f7b538e2d8/raw/0067c9b32f60362c74f4c445a080beed06b07eb3/VGG_ILSVRC_16_layers_deploy.prototxt

bklare:
@caotto It seems like enrolling a single image would take a long time b/c it needs to pool up a ton of images before passing into the net

caotto:
The net definition that you are using doesn't include a memory data layer, so I think this wrapper will not work with it. 

You might try substituting the data part in the net definition with a memory data layer (I think Jordan can give you a working example).

JordanCheney:
@bklare 2 things should help fix this. The basic issue is that OpenBR requires an explicit MemoryDataLayer to be the first layer in the net (this is not well documented ATM). You are most likely segfaulting because the first layer is not a MemoryDataLayer and line 100 calls a function (AddMatVector) that isn't defined in the first layer you have in your model file. I would replace the first 6 lines of your model file with this:

    name: "Brendan's test net"
    layer {
        name: "data"
        type: "MemoryData"
        top: "data"
        top: "label"
        memory_data_param {
            batch_size: B
            channels: 3
            width: X
            height: Y
        }
    }

I left some of the variables blank. If you just want to enroll a single image I would keep batch size at 1. You lose some advantages of batch processing (like higher throughput) but the implementation is much simpler.

The second point is that the CaffeFV transform is set up to output a feature vector (that's the FV in the name). Your net is currently functioning as a classifier, each node in the output layer represents a specific class. I have a second Caffe transform coming up that handles this use case (net as a classifier instead of feature descriptor). This won't break anything but you should be aware of it.

openbr/plugins/classification/caffe.cpp
```diff
@@ -0,0 +1,116 @@
...
+    bool timeVarying() const
+    {
+        return gpuDevice < 0 ? false : true;
+    }
+
+    void project(const Template &src, Template &dst) const
+    {
+        CaffeNet *net = caffeResource.acquire();
+
+        MemoryDataLayer<float> *data_layer = static_cast<MemoryDataLayer<float> *>(net->layers()[0].get());
```
caotto: @JordanCheney: if you do dynamic cast here, will it fail properly when layer 0 is not a MemoryDataLayer?

JordanCheney:
No :( dynamic_cast gives the save behavior as static in this case. It looks like I can query the layer type from the prototxt file, so I'll add a check that it equals "MemoryData".

caotto:
dynamic cast doesn't give you a null pointer? how odd

caotto:
like it's expected that static_cast doesn't work (i.e. completes the cast,
but actually the resulting pointer is bad), but I thought this is what
dynamic_cast is supposed to be for

JordanCheney:
dynamic cast does give a null pointer but it doesn't throw an error. So with static_cast or dynamic_cast a segfault is thrown at line 100 (when we try and access the layer). Do you think it's more elegant to check for a null pointer or to check against the type of the layer?

caotto:
Checking the type string is OK, basically for future reference, for
downcasts like this, static_cast will always appear to work (since it
doesn't check what the actual class of the pointer being cast is, just that
the cast is ever valid), but dynamic cast will give a NULL pointer unless
the cast is correct (i.e. if you are downcasting to the right sub-class for
the object being cast). Dynamic cast is also slower

caotto:
@JordanCheney  Anyway, it's worth pushing these updates (inheriting from UntrainableMetaTransform, and failing properly on networks with no memory data layer) when you have the time

bklare:
...and thank you both for the help before. I am bouncing back and forth on
a few tasks. Your comments helped a lot.

JordanCheney:
@caotto I'll tag you in a new PR for the changes

PR #392, opened 2015-07-01
Utility Update

sklum:
This branch moves some QtUtils to be exported under GUI utils because I've found them useful in that scope.  @jklontz?

### DIFF
--- a/openbr/core/qtutils.h
+++ b/openbr/core/qtutils.h
@@ -34,8 +34,6 @@
 namespace QtUtils
 {
     /**** File Utilities ****/
-    QStringList getFiles(QDir dir, bool recursive);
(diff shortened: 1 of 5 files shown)


openbr/core/qtutils.h
```diff
@@ -34,8 +34,6 @@
 namespace QtUtils
 {
     /**** File Utilities ****/
-    QStringList getFiles(QDir dir, bool recursive);
```
jklontz: should there be some deleted code in qtutils.cpp?

sklum: sorry, missed that :smile:

jklontz:
lgtm

PR #393, opened 2015-07-07
Option to write labels to the matlab gallery

bklare:


jklontz:
Generally plugins for which you are the owner of don't need to go through code review if you don't want to. Having said that, lgtm.

bklare:
Noted and thanks!

Issue #390, opened 2015-06-28
Plotting


caotto:
I am making a couple figures using -plot, and I would describe the experience as infuriating. I am using this issue to document the problems that I have with it:

-plot outputs a bunch of figures that I have no interest in. For example, I want to plot a single CMC curve, and I get a 8 page pdf document. Later, I will want to plot a single ROC curve, and will also get an 8 page document

I want to plot a CMC, up to rank 20, and don't want a log scale on the x-axis. This is not possible, because even though cmcOptions exists, it only supports a small list of things, and adding things to it requires 1. knowing how to use ggplot, and knowing how to insert things into the R string building exercise being carried out in Plot.

I think generating .R files via string building is probably crazy. The code that writes these things looks like a mess, and we'd probably be better off having R functions that take arguments, rather than doing all this string building

y axis ticks for non-log scales are at 25% increments. This should be 10%

bklare:
Charles - just a few inline thoughts below bc while there is without
question room for improvement, I think @bklein has added a lot of improved
functionality recently. Perhaps we can narrow down from here the ones that
are most needed after others comment as well, and then add ticket(s) in
github for them.

On Jun 28, 2015 2:20 PM, "caotto" <notifications@github.com> wrote:
>
> I am making a couple figures using -plot, and I would describe the
experience as infuriating. I am using this issue to document the problems
that I have with it:
>
> -plot outputs a bunch of figures that I have no interest in. For example,
I want to plot a single CMC curve, and I get a 8 page pdf document. Later,
I will want to plot a single ROC curve, and will also get an 8 page document

I do not understand the preferred solution as everyone desires different
metrics. Improvements were made before for the plots to be formatted such
that they embed in a more aesthetic manner. In terms of multi page plots,
what is the issue with grabbing the page of interest and using it as a
figure?

>
> I want to plot a CMC, up to rank 20, and don't want a log scale on the
x-axis. This is not possible, because even though cmcOptions exists, it
only supports a small list of things, and adding things to it requires 1.
knowing how to use ggplot, and knowing how to insert things into the R
string building exercise being carried out in Plot.

I agree that it would be nice for command line options to exist for setting
different parameters.

>
> I think generating .R files via string building is probably crazy. The
code that writes these things looks like a mess, and we'd probably be
better off having R functions that take arguments, rather than doing all
this string building

I think better command line options should make this less of an issue as
the R code should not need to be edited *if* sufficient flexibility exists.
[...]

PR #395, opened 2015-07-07
Plotting customization

bhklein:
This PR addresses some of the issues raised in #390.

* Added separate plot options for DET, and IET
* Set default ticks to 10% increments.
* Allow setting of x-axis limits
* Specify log/linear axis
* Added option to plot first two metadata pages
* Added option to output metadata tables to csv files
* Updated plotting docs

### DIFF
--- a/docs/docs/api_docs/c_api/functions.md
+++ b/docs/docs/api_docs/c_api/functions.md
@@ -595,14 +595,15 @@ In order of their output, the figures are:
 1. Metadata table
 2. Receiver Operating Characteristic (ROC)
 3. Detection Error Tradeoff (DET)
-4. Score Distribution (SD) histogram
-5. True Accept Rate Bar Chart (BC)
-6. Cumulative Match Characteristic (CMC)
-7. Error Rate (ERR) curve
+4. Identification Error Tradeoff (IET)
+5. Cumulative Match Characteristic (CMC)
+6. Score Distribution (SD) histogram
+7. True Accept Rate Bar Chart (BC)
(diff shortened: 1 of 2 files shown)


bklare:
Just ran the csv portion of this code. Perfect. Thank you @bhklein !

bhklein:
glad to hear @bklare

jklontz:
@caotto can you review?

caotto:
What is the current status of windows support for the project? I think there was some talk about dropping support for VS2012, has that happened already?

caotto:
@jklontz ping

caotto:
Anyway the build being broken on windows/vs2012 is more or less unrelated to this PR I guess. 

Regardless of future plans, uniform handling of plot options is a good improvement.

Some examples of setting various options in the documentation would be a good idea, e.g. I don't know what the syntax for setting breaks/ticks is, users in general may not know offhand how a QPointF is expected to be specified.

jklontz:
@caotto I'd like to move to MSVC2013, but as you're the only windows developer at this point, you're welcome to make the call.

caotto:
Well to speak to my immediate concern, does that mean the project will build if I do install VS2013, or just that no one is trying to maintain windows compatibility

jklontz:
I think that currently no one tracks the build status on windows, so we don't know when we break windows compatibility.

bhklein:
docs updated

caotto:
:+1:

bhklein:
thanks

PR #400, opened 2015-07-17
ope

privadocs:


jklontz:
What's the deal here? I think these commits already landed in master...

PR #404, opened 2015-07-27
DLandmarker with no rect specified

JordanCheney:
@sklum can you review?

I am running into a edge case where I need to pass a cropped face into the DLandmarker transform. The transform currently requires a face rect and an image as input, so this introduces a case where if no rect is present the entire image is considered a rect.


sklum:
@JordanCheney lgtm

PR #407, opened 2015-08-04
Caffe classifier

JordanCheney:
@caotto. This has 2 parts, 1) Changes we talked about in the previous PR, namely change to MetaTransform and check if the first layer is a MemoryDataLayer. 2) A second caffe classifier transform. CaffeFV assumes output is a feature vector, CaffeClassifier assumes output is a score distribution for the input belonging to a range of classes. It finds the maximum score of the output nodes and stores the index as the label and the score as the confidence.

### DIFF
--- a/openbr/plugins/classification/caffe.cpp
+++ b/openbr/plugins/classification/caffe.cpp
@@ -93,7 +93,7 @@ class CaffeFVTransform : public UntrainableMetaTransform
         if (net->layers()[0]->layer_param().type() != "MemoryData")
             qFatal("OpenBR requires the first layer in the network to be a MemoryDataLayer");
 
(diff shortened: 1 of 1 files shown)


caotto:
Generally looks OK, there is some marginal code duplication here (all properties are identical between the two transforms, the new transform is duplicate code up to about line 168). 

You could reduce this via refactoring: put project, up to 168 in a seaprate funciton, which returns the output blob (excluding the dst = src part of course), then either inherit the classification transform from the feature vector transform, or inherit both from a common subclass, keeping the property definitions in the base class.

JordanCheney:
How does it look now?

caotto:
Looks better in terms of source code. Are inherited properties handled in terms of documentation? If no, probably good to copy the property documentation to the sub-classes

JordanCheney:
They aren't handled in the docs so I added them to each class

caotto:
looks good to me

PR #408, opened 2015-08-11
Embedded

sklum:
This branch has fixes to enable building with `BR_EMBEDDED` and moves the utilities that were in `openbr/gui/utility.*` but weren't GUI specific `openbr/core/utility.*`.  @jklontz tap tap.

### DIFF
--- a/openbr/plugins/gallery/xml.cpp
+++ b/openbr/plugins/gallery/xml.cpp
@@ -14,9 +14,7 @@
  * limitations under the License.                                            *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
-#ifndef BR_EMBEDDED
 #include <QtXml>
-#endif // BR_EMBEDDED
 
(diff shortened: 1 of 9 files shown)


openbr/plugins/gallery/xml.cpp
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.                                            *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+#ifndef BR_EMBEDDED
```
jklontz: xmlGallery won't work without QtXml, so why not just one `#ifndef` around the whole file?

jklontz:
lgtm otherwise

openbr/plugins/gallery/xml.cpp
sklum: I would have done it that way, but `dbGallery` deals with `BR_EMBEDDED` this way so I assumed there was a reason for it.

PR #409, opened 2015-08-11
Network

sklum:
Even if one builds with `BR_EMBEDDED`, OpenBR still depends on the `QtNetwork` library.  This branch fixes that.  @jklontz?

### DIFF
--- a/openbr/openbr.cpp
+++ b/openbr/openbr.cpp
@@ -303,6 +303,7 @@ const char *br_version()
     return version.data();
 }
 
+#ifdef BR_WITH_QTNETWORK 
 void br_slave_process(const char *baseName)
 {
     WorkerProcess *worker = new WorkerProcess;
@@ -311,6 +312,7 @@ void br_slave_process(const char *baseName)
     worker->mainLoop();
     delete worker;
 }
+#endif
 
 br_template br_load_img(const char *data, int len)
 {
--- a/openbr/plugins/openbr_internal.h
+++ b/openbr/plugins/openbr_internal.h
@@ -366,6 +366,7 @@ protected:
     CompositeTransform() : TimeVaryingTransform(false) {}
 };
 
+#ifdef BR_WITH_QTNETWORK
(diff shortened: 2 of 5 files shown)


openbr/openbr.cpp
```diff
@@ -303,6 +303,7 @@ const char *br_version()
     return version.data();
 }
 
+#ifdef BR_WITH_QTNETWORK 
 void br_slave_process(const char *baseName)
 {
```
jklontz: Maybe have br_slave_process error out without QtNetwork so you don't also have to check for it in `br.cpp`?

openbr/plugins/openbr_internal.h
```diff
@@ -366,6 +366,7 @@ class BR_EXPORT CompositeTransform : public TimeVaryingTransform
     CompositeTransform() : TimeVaryingTransform(false) {}
 };
 
+#ifdef BR_WITH_QTNETWORK
```
jklontz: Strictly speaking, this doesn't seem necessary.

sklum: `mainLoop` is defined in `plugins/processwrapper.cpp` which cannot be compiled without `QtNetwork`.  Compilation fails without this `ifdef` block.

caotto: due to undefined reference? you could also do an
#if network
real implementation
#else
dummy implementation
#endif
structure

sklum: Yes.  The real implementation requires the definition of `EnrollmentWorker` so it has to live in `processwrapper.cpp` right?  Is there a severe issue with what I have right now?

caotto: I mean, you can put an #if, else, #endif in processwrapper.cpp (compiling the real implementation if network is defined, otherwise compiling a stub implementation), I don't care that much but you could touch fewer files that way

jklontz:
I think this issue has been fixed now.

PR #427, opened 2015-09-21
Fixed #414

### LINKED ISSUE #414: br command gives error
After following installation instructions on openbiometrics.org, tested br in terminal on Mac and received following error.
cmake version: 2.8.12 (also tried with version 3.0.2)
Tried without success setting each RPATH and LD_LIBRARY_PATH to /usr/local/lib:/Users/m30039/Qt5.4.1/5.4/clang_64/lib

$ br
dyld: Library not loaded: @rpath/QtConcurrent.framework/Versions/5/QtConcurrent
  Referenced from: /usr/local/bin/br
  Reason: image not found
Trace/BPT trap: 5

mmtaborsky:
This fixes the linker error in issue #414. @jklontz can you look at this?


bklare:
Tiny T in the hizzie!!!

sklum:
@mmtaborsky Looks good to me - nice catch.  We may want to go down the path (ha pun) of setting the `RPATH` to be `@loaderPath/.;@loaderPath/../lib/` or something of that sort when `BR_INSTALL_DEPENDENCIES` is turned on.  Having said that, I think this should be merged.

jklontz:
ty

PR #433, opened 2015-10-06
Detection api update

JordanCheney:
@sklum I think this is ready for review. The purpose of this is to facilitate representations that use metadata as well as the image.


openbr/core/boost.cpp
```diff
@@ -131,23 +131,17 @@ static CvMat* cvPreprocessIndexArray( const CvMat* idx_arr, int data_arr_size, b
...
 //------------------------------------- FeatureEvaluator ---------------------------------------
 
-void FeatureEvaluator::init(Representation *_representation, int _maxSampleCount, int channels)
+void FeatureEvaluator::init(Representation *_representation, int _maxSampleCount)
 {
     representation = _representation;
-
-    int dx, dy;
-    Size windowSize = representation->windowSize(&dx, &dy);
-    data.create((int)_maxSampleCount, (windowSize.width + dx) * (windowSize.height + dy), CV_32SC(channels));
```
sklum: I need the ability to specify number of channels.

openbr/plugins/representation/random.cpp
```diff
@@ -40,26 +40,14 @@ class RandomRepresentation : public Representation
...
+        return representation->preprocess(src);
     }
 
-    float evaluate(const Mat &image, int idx) const
+    float evaluate(const Template &src, int idx) const
     {
-        return representation->evaluate(image,features[idx]);
-    }
-
-    Mat evaluate(const Mat &image, const QList<int> &indices) const
```
sklum: Why was this removed entirely?

openbr/plugins/representation/haar.cpp
```diff
@@ -39,64 +39,64 @@ class HaarRepresentation : public Representation
 
     void init()
     {
-        if (features.isEmpty()) {
```
sklum: This is actually important.  Init is called twice in some circumstances so the number of features you get will be doubled.

openbr/core/boost.cpp
JordanCheney: Since the entire template is stored the whole matrix is available as is to the representation so you shouldn't have to specify it here anymore

openbr/plugins/representation/random.cpp
JordanCheney: No class called evaluate with a list of indices every class just did it with a single index so I thought it could be removed

openbr/plugins/representation/haar.cpp
JordanCheney: Why is init called twice?

openbr/plugins/representation/random.cpp
JordanCheney: I'm changing my mind here. I'll leave both evaluates (single index and list of indices)

openbr/core/boost.cpp
sklum: I see.  Didn't notice that we're just creating the cls matrix now.

openbr/plugins/representation/haar.cpp
sklum: Not sure.

JordanCheney: Ok I'll add this in for all representations then

JordanCheney:
@sklum I think this has all of the changes

JordanCheney:
Ok the file is gone

sklum:
lgtm

Issue #431, opened 2015-10-04
stasm metadata are probably overwritten in -enrollAll mode


koenokatachi:
I want to run Stasm over a picture with multiple faces and outline rectangles and landmarks. So

```
br -algorithm 'Open+Cascade(FrontalFace)+Stasm+Draw(inPlace=true)' -enrollAll -enroll input.png output.png
```
Stasm landmarks are only outlined for a single face.

Sample input and output:
![home2](https://cloud.githubusercontent.com/assets/11669814/10270715/04260684-6b03-11e5-9b75-721742eb130a.png)
![test](https://cloud.githubusercontent.com/assets/11669814/10270717/0aa2ef22-6b03-11e5-827b-1b1c73231e18.png)

To be sure stasm actually sees the rest of the faces I've made the outlined face not eligible for detectors. Another pair of input and output:
![home3](https://cloud.githubusercontent.com/assets/11669814/10270724/3ec93220-6b03-11e5-9969-7ba45714683b.png)
![test3](https://cloud.githubusercontent.com/assets/11669814/10270725/43c83ed8-6b03-11e5-9bc9-eafd933cde78.png)

Enrolling into ```xml``` format i see the same Points and StasmXEye entries for each face.

It looks like stasm overwrites its metadata with each invocation.

jklontz:
Hmm yeah this is a nasty bug. @bklare @sklum Looks like the `Cascade(FrontalFace)+Stasm` combination isn't well supported currently. Without eye coordinates (searchPinned=false) STASM will re-run face detection, effectively clobbering previous face detections results and detecting the same face 3 times over in the above case. Not sure about the best way to proceed...

sklum:
It's a known issue that's been brought up on the mailing list - if someone is willing to dig through the STASM code to remove its internal detection I'd be happy to review the PR, but the STASM source is not fun to play with.

jklontz:
What about the alternative of removing `Cascade(FrontalFace)` from the algorithm string and detecting all faces from the start with Stasm?

sklum:
Oh, I see. This is an issue with the released model.  Yikes.  Let me verify that that even works as expected.

sklum:
Doesn't currently work this way, but I might be able to patch it so that it does work.

bklare:
Inefficiency from the double call to face detection aside, I do not
understand why this issue is occurring. When using enrollAll=true, after
the call to CascadeTransform, shouldn't all the detected faces be cropped
[...]

PR #435, opened 2015-10-11
Stasm properly uses internal detector to landmark all faces in image

sklum:
This branch represents a minimally invasive change solving the issue with the current release described in #431.  The correspondingly released models should work on images with multiple faces now (with similar behavior to `CascadeTransform` in terms of detection) as I have updated the algorithm string in `algorithms.cpp` as well.  I also cleaned up the `StasmTransform` a bit and removed some old code from `3rdparty`.  @jklontz thoughts?

screen shot 2015-10-11 at 5 43 35 pm
screen shot 2015-10-11 at 5 43 31 pm
screen shot 2015-10-11 at 5 43 28 pm


bklare:
Nice job @sklum !

sklum:
Realized this morning that with the stuff I exposed yesterday, I could just use the rects provided by `Cascade`.  `Stasm` now works with (i) any rects in template metadata (via an input variable or in the rect list) or (ii) if no rects are present in either of those it will try to do its own detection.  Should be ready to merge now @jklontz.

jklontz:
Do it! Thanks so much :)

PR #439, opened 2015-10-29
SlidingWindow Cleanup

jklontz:
@sklum Here's my first pass through the code. This just does the obvious stuff like hoisting invariants outside of loops, removing redundant variables, improving const correctness, etc. This reduced the runtime on my test case by 30% (from 83 to 58 seconds). Comments inline. @JordanCheney FYI.

### DIFF
--- a/openbr/plugins/imgproc/slidingwindow.cpp
+++ b/openbr/plugins/imgproc/slidingwindow.cpp
@@ -91,50 +91,44 @@ class SlidingWindowTransform : public MetaTransform
                 continue;
             }
 
+            // SlidingWindow assumes that all matricies in a template represent
+            // different channels of the same image!
+            const Size imageSize = t.m().size();
             const int minSize = t.file.get<int>("MinSize", this->minSize);
-            Size minObjectSize(minSize, minSize);
-            Size maxObjectSize;
-
-            Mat m = t.m();
             QList<Rect> rects;
             QList<float> confidences;
 
-            if (maxObjectSize.height == 0 || maxObjectSize.width == 0)
-                maxObjectSize = m.size();
+            int dx, dy;
+            const Size originalWindowSize = classifier->windowSize(&dx, &dy);
 
             for (double factor = 1; ; factor *= scaleFactor) {
-                int dx, dy;
-                Size originalWindowSize = classifier->windowSize(&dx, &dy);
-
-                Size windowSize(cvRound(originalWindowSize.width*factor), cvRound(originalWindowSize.height*factor) );
-                Size scaledImageSize(cvRound(m.cols/factor ), cvRound(m.rows/factor));
-                Size processingRectSize(scaledImageSize.width - originalWindowSize.width, scaledImageSize.height - originalWindowSize.height);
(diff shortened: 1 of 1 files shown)


openbr/plugins/imgproc/slidingwindow.cpp
```diff
@@ -91,50 +91,44 @@ class SlidingWindowTransform : public MetaTransform
...
-                Size windowSize(cvRound(originalWindowSize.width*factor), cvRound(originalWindowSize.height*factor) );
-                Size scaledImageSize(cvRound(m.cols/factor ), cvRound(m.rows/factor));
-                Size processingRectSize(scaledImageSize.width - originalWindowSize.width, scaledImageSize.height - originalWindowSize.height);
+                const Size windowSize(cvRound(originalWindowSize.width*factor), cvRound(originalWindowSize.height*factor));
+                const Size scaledImageSize(cvRound(imageSize.width/factor), cvRound(imageSize.height/factor));
+                const Size processingRectSize(scaledImageSize.width - originalWindowSize.width, scaledImageSize.height - originalWindowSize.height);
 
                 if (processingRectSize.width <= 0 || processingRectSize.height <= 0)
                     break;
-                if (windowSize.width > maxObjectSize.width || windowSize.height > maxObjectSize.height)
```
jklontz: It seems that this condition is impossible to reach as the one above will always be true first?

openbr/plugins/imgproc/slidingwindow.cpp
```diff
@@ -91,50 +91,44 @@ class SlidingWindowTransform : public MetaTransform
...
-                    scaleBuffer.append(scaledImage);
+                    resize(m, scaledImage, scaledImageSize, 0, 0, CV_INTER_LINEAR);
+                    rep.append(scaledImage);
                 }
+                rep = classifier->preprocess(rep);
 
-                Template rep(t.file);
-                rep = classifier->preprocess(scaleBuffer);
+                // Pre-allocate the window to avoid constructing this every iteration
+                Template window(t.file);
```
jklontz: While I didn't test this optimization in isolation, this is probably the primary source of speedup.

openbr/plugins/imgproc/slidingwindow.cpp
```diff
@@ -155,12 +149,12 @@ class SlidingWindowTransform : public MetaTransform
...
 
             if (!enrollAll && rects.empty()) {
-                rects.append(Rect(0, 0, m.cols, m.rows));
+                rects.append(Rect(0, 0, imageSize.width, imageSize.height));
                 confidences.append(-std::numeric_limits<float>::max());
             }
 
             for (int j=0; j<rects.size(); j++) {
-                Template u(t.file, m);
+                Template u = t;
```
jklontz: I assume you meant to propagate all the matrices?

sklum:
lgtm

openbr/plugins/imgproc/slidingwindow.cpp
sklum: Yup.

PR #443, opened 2015-11-08
Option to output detected regions as rect list instead of seperate templates

bklare:
@sklum @JordanCheney FYI. Let me know if you have any issues with this change.


sklum:
lgtm @bklare

JordanCheney:
@bklare looks good to me as well. Keep in mind some formats (like csv) require you to explicitly store the rectangles in the metadata (the rects field isn't serialized). This won't work in those cases

jklontz:
Closed due to inactivity

PR #444, opened 2015-11-10
remove randomlib

jklontz:
@sklum @caotto This pull requests removes the RandomLib dependency which doesn't compile on android due to `gethostid` not existing (among other issues). Thankfully there are only two places in the code base where we use this library, so removing it appears pretty straightforward. Comments inline. @bklare FYI.


caotto:
Did you check that training on windows still works (produces the same accuracy as on linux)

jklontz:
Here's the diff which doesn't show up well:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ae24144..d8047ea 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,10 +92,6 @@ set(BR_THIRDPARTY_LIBS ${BR_THIRDPARTY_LIBS} ${OPENCV_DEPENDENCIES})
 # Find Alphanum
 find_package(Alphanum REQUIRED)
 
-# Add RandomLib
-set(BR_THIRDPARTY_SRC ${BR_THIRDPARTY_SRC} ${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/RandomLib-1.9/src/Random.cpp)
-include_directories(3rdparty/RandomLib-1.9/include)
-
 # Find cvmatio (if using it)
 set(BR_WITH_CVMATIO OFF CACHE BOOL "Build with cvmatio library to read Matlab data files (required to use Caltech Pedestrians dataset)")
 if(${BR_WITH_CVMATIO})
diff --git a/app/br/br.cpp b/app/br/br.cpp
index 89b96be..c421f3d 100644
--- a/app/br/br.cpp
+++ b/app/br/br.cpp
@@ -237,6 +237,9 @@ public:
             } else if (!strcmp(fun, "setHeader")) {
                 check(parc == 3, "Incorrect parameter count for 'setHeader'.");
                 br_set_header(parv[0], parv[1], parv[2]);
+            } else if (!strcmp(fun, "srand")) {^M
+                check(parc == 1, "Incorrect parameter count for 'srand'.");^M
+                srand(atoi(parv[1]));^M
             } else if (!strcmp(fun, "br")) {
                 printf("That's me!\n");
             } else if (parc <= 1) {
@@ -305,7 +308,8 @@ private:
                "-version\n"
                "-daemon\n"
                "-slave\n"
-               "-exit\n");
+               "-exit\n"^M
+               "-srand <int>\n");^M
     }
 };
 
diff --git a/openbr/core/common.cpp b/openbr/core/common.cpp
index c3802e4..173dfcc 100644
--- a/openbr/core/common.cpp
+++ b/openbr/core/common.cpp
@@ -16,30 +16,17 @@
 
 #include "common.h"
 #include <QMutex>
-#include <RandomLib/Random.hpp>
 
 using namespace std;
 
-static RandomLib::Random g_rand;
 static QMutex rngLock;
 
 /**** GLOBAL ****/
-void Common::seedRNG() {
-    QMutexLocker lock(&rngLock);
-
-    static bool seeded = false;
-    if (!seeded) {
-        srand(0); // We seed with 0 instead of time(NULL) to have reproducible randomness
-        seeded = true;
-        g_rand.Reseed(0);
-    }
-}
-
 double Common::randN()
 {
+    // The function rand() is not reentrant or thread-safe, since it uses hidden state that is modified on each call.
     QMutexLocker lock(&rngLock);
-
-    return g_rand.FloatN();
+    return (double)rand() / RAND_MAX;
 }
 
 QList<int> Common::RandSample(int n, int max, int min, bool unique)
diff --git a/openbr/core/common.h b/openbr/core/common.h
index 843dc96..5b078fa 100644
--- a/openbr/core/common.h
+++ b/openbr/core/common.h
@@ -229,7 +229,6 @@ double randN();
  * If unique then there will be no repeated integers.
  * \note Algorithm is inefficient for unique vectors where n ~= max-min.
  */
-void seedRNG();
 QList<int> RandSample(int n, int max, int min = 0, bool unique = false);
 QList<int> RandSample(int n, const QSet<int> &values, bool unique = false);
 
diff --git a/openbr/openbr_plugin.cpp b/openbr/openbr_plugin.cpp
index 67783b3..5fa7765 100644
--- a/openbr/openbr_plugin.cpp
+++ b/openbr/openbr_plugin.cpp
@@ -29,7 +29,6 @@
 #include <QtConcurrentRun>
 #include <algorithm>
 #include <iostream>
-#include <RandomLib/Random.hpp>
 
 #ifndef BR_EMBEDDED
 #include <QApplication>
@@ -661,16 +660,12 @@ QList<int> TemplateList::applyIndex(const QString &propName, const QHash<QString
     return result;
 }
 
-TemplateList TemplateList::partition(const QString &inputVariable, unsigned int randomSeed, bool overwrite) const
+TemplateList TemplateList::partition(const QString &inputVariable, bool random, bool overwrite) const
 {
     const int crossValidate = std::abs(Globals->crossValidate);
     if (crossValidate < 2)
         return *this;
 
-    // Use separate RNG from Common::randN() to avoid re-seeding the global RNG.
-    // rng is seeded with the inputVariable hash in order to maintain partition across br runs, given the same randomSeed seed.
-    RandomLib::Random rng;
-
     TemplateList partitioned = *this;
 
     if (Globals->verbose)
@@ -690,10 +685,8 @@ TemplateList TemplateList::partition(const QString &inputVariable, unsigned int
         } else {
             if (partitioned[i].file.contains(inputVariable)) {
                 const QByteArray md5 = QCryptographicHash::hash(partitioned[i].file.get<QString>(inputVariable).toLatin1(), QCryptographicHash::Md5);
-                if (randomSeed) {
-                    quint64 labelSeed = md5.toHex().right(8).toULongLong(0, 16) + randomSeed;
-                    rng.Reseed(labelSeed);
-                    partitioned[i].file.set("Partition", rng.Integer() % crossValidate);
+                if (random) {
+                    partitioned[i].file.set("Partition", rand() % crossValidate);
                 } else if (!partitioned[i].file.contains("Partition") || overwrite) {
                     // Select the right 8 hex characters so that it can be represented as a 64 bit integer without overflow
                     partitioned[i].file.set("Partition", md5.toHex().right(8).toULongLong(0, 16) % crossValidate);
@@ -1305,7 +1298,8 @@ void br::Context::initialize(int &argc, char *argv[], QString sdkPath, bool useG
     // The message handler requires a valid `Globals` so we set it after `Globals` is constructed
     qInstallMessageHandler(messageHandler);
 
-    Common::seedRNG();
+    // We seed with 0 instead of time(NULL) to have reproducible randomness
+    srand(0);
 
     // Trigger registered initializers
     QList< QSharedPointer<Initializer> > initializers = Factory<Initializer>::makeAll();
diff --git a/openbr/openbr_plugin.h b/openbr/openbr_plugin.h
index 783cb5d..5c227e5 100644
--- a/openbr/openbr_plugin.h
+++ b/openbr/openbr_plugin.h
@@ -319,7 +319,7 @@ struct TemplateList : public QList<Template>
     BR_EXPORT static TemplateList relabel(const TemplateList &tl, const QString &propName, bool preserveIntegers);
 
     /*!< \brief Assign templates to folds partitions. */
-    BR_EXPORT TemplateList partition(const QString &inputVariable, unsigned int randomSeed = 0, bool overwrite = false) const;
+    BR_EXPORT TemplateList partition(const QString &inputVariable, bool random = false, bool overwrite = false) const;
 
     BR_EXPORT QList<int> indexProperty(const QString &propName, QHash<QString, int> * valueMap=NULL,QHash<int, QVariant> * reverseLookup = NULL) const;
     BR_EXPORT QList<int> indexProperty(const QString &propName, QHash<QString, int> &valueMap, QHash<int, QVariant> &reverseLookup) const;
```

jklontz:
@caotto The only potentially affected plugin appears to be `RandomSubspace` which isn't used in any production algorithm a.t.m. Given how we seed and lock RNG access, I'd be floored if an issue still exists on Windows...

sklum:
Fine by me.

jklontz:
@caotto Does ffc4297278315d35c6725b057c9696c9f0be5898 adequately address you concern?

caotto:
I am not in a position to test this at the moment, but disabling parallel training should address the issue

jklontz:
Any objection to merging?

caotto:
go ahead

PR #447, opened 2015-11-12
EvalDetection change.

sklum:
Updates EvalDetection to ignore predicted bounding boxes associated with ground truth boxes that are beyond the specified thresholds.  @JordanCheney looks like the logic for this was already there but not being used.


JordanCheney:
This looks good.

PR #449, opened 2015-11-17
Commented unnecessary qDebug msg

DarwinsBuddy:
This message is not at all helpful and is very tricky to get rid of via CMake.
The message is spamming stdout unnecessarily.


jklontz:
This message generally indicates an incorrect use of the API, where a `TemplateList` (of length 1) should be used instead of a `Template`.

PR #450, opened 2015-12-08
fix compile error with dlib

mgrewe:
Hey guys,

just a short fix to make your awesome framework compile with dlib again (shape_predictor was missing).

Thanks and Cheers
Martin

### DIFF
--- a/openbr/plugins/classification/dlib.cpp
+++ b/openbr/plugins/classification/dlib.cpp
@@ -1,5 +1,5 @@
 #include <opencv2/imgproc/imgproc.hpp>
-#include <dlib/image_processing/frontal_face_detector.h>
+#include <dlib/image_processing.h>
 #include <dlib/opencv.h>
 
 #include "openbr/plugins/openbr_internal.h"


jklontz:
I'm not opposed to merging this, but we build with dlib all the time without issue (`dlib-18.16` that's downloaded and built automatically with `BR_WITH_DLIB=ON`), and `frontal_face_detector.h` appears to include `../image_processing.h`. Is there something else going on with your build?

mgrewe:
Dear Josh,

thanks for your answer. Yes, you are right. Dlib guys started to clean up includes in image_processing as of 18.17 (see http://sourceforge.net/p/dclib/code/ci/5eb8a31bfc198c84a0b56ce68601c70fb094ffa0/). Unfortunately, there is some mess with missing STL includes when only adding '#include <image_processing/shape_predictor.h'. I see no reason why not just including the main package include '#include <image_processing.h>'. I changed my commit accordingly.

Martin

jklontz:
Sorry for the confusion, I didn't mean to suggest you should change it. I don't think doing just `#include <image_processing.h>` will work with dlib 18.16. Anyway, I applied your original fix: https://github.com/biometrics/openbr/commit/6e4b645cf101b0615b80c7df0afcb031335c6b0a

PR #464, opened 2016-03-31
CUDA-accelerated plugin support

DepthDeluxe:
This pull requests adds some CUDA accelerated functions to OpenBR.  This gives vast performance improvements over existing implementation of algorithms.  A good place to start understanding our implementation is in `openbr/plugins/cuda/README.md`.

### Where new stuff is
`openbr/plugins/cuda` - new CUDA accelerated plugins
`openbr/plugins/cuda/README.md` - README that digs into more detail of how this is implemented
`openbr/plugins/cuda/cudadefines.hpp` - contains common defines related to this folder

### Building with/without CUDA support
To enable CUDA plugins, run cmake with the option `-DBR_WITH_CUDA=ON`.  You must have the NVidia CUDA compiler to build and a CUDA compatible graphics card to run CUDA plugins. This defaults to off.

### New Plugin List
This is the list of the new plugins in this PR.  They all reside in the `openbr/plugins/cuda` folder.

1. `CUDACopyFrom` - Copy data from graphics card to BR Template object on CPU
1. `CUDACopyTo` - Copy data to graphics card from BR Template object on CPU
1. `CUDAAffine` - CUDA accelerated affine transform
1. `CUDACvtFloat` - Convert byte image data to floating point form on GPU
1. `CUDAL2` - CUDA accelerated L2 distance computation
1. `CUDALBP` - CUDA accelerated local binary patterns
1. `CUDAPCA` - CUDA accelerated principal component analysis (training still on CPU)
1. `CUDARGB2Grayscale` - Converts RGB images on graphics card to grayscale


CMakeLists.txt
```diff
@@ -106,7 +106,7 @@ endif()
 
 # Compiler flags
 if(UNIX)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wno-strict-overflow -Wno-comment -Wno-unknown-pragmas -fvisibility=hidden -fno-omit-frame-pointer")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wno-strict-overflow -Wno-comment -Wno-unknown-pragmas -fvisibility=hidden -fno-omit-frame-pointer -pthread")
```
jklontz: What's the deal here?

openbr/CMakeLists.txt
```diff
@@ -28,18 +29,30 @@ if(NOT BR_EMBEDDED)
...
-endif()
+# add CUDA compiler flags if CUDA support is enabled
+option(BR_WITH_CUDA "Build CUDA-accelerate plugins." OFF)
+if (BR_WITH_CUDA)
+  FIND_PACKAGE(CUDA REQUIRED)
+  SET(CUDA_SEPARABLE_COMPILATION ON)
+  SET(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS} -arch=sm_20;--compiler-options "-fPIC")
 
-add_library(openbr SHARED ${SRC} ${BR_CORE} ${BR_JANUS} ${BR_GUI} ${BR_ICONS} ${BR_THIRDPARTY_SRC} ${BR_RESOURCES} ${NATURALSTRINGCOMPARE_SRC} ${THIRDPARTY_RESOURCES})
+  cuda_add_library(openbr SHARED ${SRC} ${BR_CORE} ${BR_JANUS} ${BR_GUI} ${BR_ICONS} ${BR_THIRDPARTY_SRC} ${BR_RESOURCES} ${NATURALSTRINGCOMPARE_SRC} ${THIRDPARTY_RESOURCES})
```
jklontz: Can we factor out the cuda plugins to a static library to avoid `cuda_add_library(openbr)`?

openbr/CMakeLists.txt
```diff
@@ -28,18 +29,30 @@ if(NOT BR_EMBEDDED)
...
 qt5_use_modules(openbr ${QT_DEPENDENCIES})
 set_target_properties(openbr PROPERTIES
                              DEFINE_SYMBOL BR_LIBRARY
                              VERSION ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}
                              SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR})
 target_link_libraries(openbr ${BR_THIRDPARTY_LIBS})
+
+# include CUDA dependencies
+if (BR_WITH_CUDA)
+  target_link_libraries(openbr ${CUDA_LIBRARIES})
```
jklontz: Append `CUDA_LIBRARIES` to `BR_THIRDPARTY_LIBS` instead.

openbr/plugins/cuda/cudaaffine.cpp
```diff
@@ -0,0 +1,191 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * Copyright 2012 The MITRE Corporation                                      *
```
jklontz: Assign yourself the copyright for all of these new files, but please keep the Apache 2 license.

openbr/plugins/plugins.cmake
```diff
@@ -12,6 +12,12 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${BR_THIRDPARTY_PLUGINS_DIR})
 file(GLOB SUBFILES plugins/*)
 foreach(FILE ${SUBFILES})
   if(IS_DIRECTORY ${FILE})
+    # don't add the CUDA directory in plugins folder if we have disabled CUDA
+    get_filename_component(FILE_NAME ${FILE} NAME)
```
jklontz: The preferred way to do this is add a `CMakeLists.txt` to the CUDA directory which and control the inclusion / exclusion of plugins from there. I assume this wasn't done due to how `cuda_add_library` is used?

jklontz:
I think this very nicely put together. Looking forward to resolving the few minor issues mentioned.

openbr/CMakeLists.txt
DepthDeluxe: OK I tried to do this a while back and the big problem is that if we add the CUDA plugins to the transform registry you will always have to link the CUDA library or else you will get unresolved function dependencies when linking your final program.  For example, all the Janus tests would not compile correctly when BR is built in this way.

I guess the way to get around this is to have a second CUDA plugin registry but I haven't looked into how hard that would be to implement.

openbr/plugins/plugins.cmake
DepthDeluxe: Your assumption is correct.  The `cuda/` folder is just like every other plugin folder.

It's sounding like it may actually be better if we have a `openbr/cuda-plugins` folder that is separate from the rest of the plugins.  If we do it this way, the sources will be far apart from one another which should keep the build process separate.

That way we can also put different plugin types in different folders (i.e. `cuda-plugins/imgprog`, `cuda-plugins/distance`,etc...).  If more CUDA functions are added to that one folder is going to get messy.

openbr/CMakeLists.txt
jklontz: Perhaps I don't understand the problem entirely, but I think you should be able to use `CUDA_COMPILE` on just the `.cu` files. Then append the resulting object files to `BR_THIRDPARTY_SRC`.

DepthDeluxe: @jklontz so i've been struggling to get something ideal here.  It appears that you cannot have object files in `${BR_THIRDPARTY_SRC}` because `add_library` only allows for source files.

I tried to build the files into a new static library called openbr-cuda but it appears that the plugins don't get registered into the transform registry if done that way.

The following is what I'm using:

```cmake
# add WITH_CUDA option
option(BR_WITH_CUDA "Build CUDA-accelerated plugins." OFF)

# only build this module if explicitly OK'ed
if(BR_WITH_CUDA)
  message(STATUS "Building with CUDA Support")
  find_package(CUDA REQUIRED)
  #set(CUDA_SEPERABLE_COMPILATION ON)

  set(CUDA_SRC_DIR ${PROJECT_SOURCE_DIR}/openbr/plugins/cuda)

  # configure the compiler
 set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS} -arch=sm_20;--compiler-options)

  # glob the source together
  file(GLOB CUDA_SRC ${CUDA_SRC_DIR}/*.cpp ${CUDA_SRC_DIR}/*.cu)
  message(STATUS "CUDA_SRC" ${CUDA_SRC})

  cuda_add_library(openbr-cuda STATIC ${CUDA_SRC})
  qt5_use_modules(openbr-cuda ${QT_DEPENDENCIES})
  target_link_libraries(openbr-cuda ${CUDA_LIBRARIES} ${BR_THIRDPARTY_LIBS})

  # add the compiled source and libs into the build system
  #set(BR_THIRDPARTY_SRC ${BR_THIRDPARTY_SRC})
  set(BR_THIRDPARTY_LIBS ${BR_THIRDPARTY_LIBS} openbr-cuda)

endif()
```

Do you have any recommendations for an alternative approach?

jklontz: Can you elaborate on the `add_library` approach not working? My understanding is that you can add object files, perhaps you aren't setting the `EXTERNAL_OBJECT` property for these object files?

https://cmake.org/cmake/help/v3.0/command/set_source_files_properties.html
https://cmake.org/cmake/help/v3.0/prop_sf/EXTERNAL_OBJECT.html

DepthDeluxe: Didn't know that was a thing... Added the property and everything builds together!

openbr/CMakeLists.txt
```diff
@@ -28,18 +29,15 @@ if(NOT BR_EMBEDDED)
   install(FILES ${HEADERS} DESTINATION include/openbr/gui)
 endif()
 
-# Compile third party resources
```
jklontz: Why was this deleted?

openbr/plugins/cuda/threshold.cpp
```diff
@@ -0,0 +1,44 @@
+#include <iostream>
```
jklontz: needs a copyright

openbr/plugins/distance/dist.cpp
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.                                            *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
```
jklontz: Seems like the changes to this file can be reverted

openbr/plugins/plugins.cmake
```diff
@@ -12,7 +12,15 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${BR_THIRDPARTY_PLUGINS_DIR})
 file(GLOB SUBFILES plugins/*)
 foreach(FILE ${SUBFILES})
   if(IS_DIRECTORY ${FILE})
-    set(BR_PLUGINS_DIR ${BR_PLUGINS_DIR} ${FILE})
+    # check to see if there is a cmake file for the folder, if there is, then that
```
jklontz: This is great.

openbr/plugins/cuda/cudaaffine.cu
```diff
@@ -0,0 +1,245 @@
...
+        affineInverse[4] = a11 * det;
+        affineInverse[5] = 0;
+        affineInverse[6] = (a21*a32 - a31*a22) * det;
+        affineInverse[7] = (a31*a12 - a11*a32) * det;
+        affineInverse[8] = (a11*a22 - a21*a12) * det;
+
+        // Move from affineTransform to gpuAffine (currently fake)
+        // double fakeAffine[6];
+        // fakeAffine[0] = affineTransform.at<double>(0, 0);
+        // fakeAffine[1] = affineTransform.at<double>(0, 1);
```
jklontz: remove these comments?

openbr/plugins/plugins.cmake
jklontz: I think this PR is missing the `module.cmake` file?

jklontz:
Looks like everything has been addressed to me. Is it ready to merge @DepthDeluxe ?

DepthDeluxe:
@jklontz unfortunately we have been running into some issues.  We identified what looked like an issue with `CUDAPCA` and then had a hard disk failure on our dev machine.  Right now I can't say whether or not the error (PCA returning garbage results) was due to (1) updating our master to the latest upstream master, (2) issue with dependencies in our build system (accidentally updated machine to `qt-5.6`), or (3) strange corrupted library errors due to disk corruption.

I'll try to get back to you this week with a resolution.

jklontz:
Ok, no rush!

DepthDeluxe:
@jklontz OK, spent the past week really nailing out test trials and validating the plugins we have.  All the plugins have now been tested with all options and compared to the CPU equivalent to ensure accuracy.  The new commits are just bug fixes.

The changes are now ready to be pulled in :grin:

jklontz:
Thanks!

PR #472, opened 2016-04-27
Added the option -C to curl to auto-continue downloading from the point it stopped.

StayerX:
Also added `-p` for `mkdir` for the script `downloadMEDS.sh`


jklontz:
Looks like you'll have to rebase before I can merge.

StayerX:
@jklontz , thank you. I think that now the issue has been resolved.  [Use "-C -"](http://linux.about.com/od/commands/l/blcmdl1_curl.htm) to tell curl to automatically find out where/how to resume the transfer. It then uses the given output/input files to figure that out. This should reduce the time and the trouble for anyone trying to test `openbr`.

jklontz:
Thanks!

PR #477, opened 2016-06-05
(Document) Windows Install:  Build OpenCV without CUDA

prince-0203:
It takes a long time to build OpenCV with CUDA.
Maybe it's not necessary to use CUDA on OpenBR.


jklontz:
Thanks!

PR #480, opened 2016-06-07
Fix multiprocess support

caotto:
Per discussion in 
https://github.com/biometrics/openbr/commit/56b776899118a15aaaf85810d53f15c1a2de3fcc
multiprocess support was broken by the removal of br_slave_process, and there was no deep reason why this was necessary.
This branch re-introduces it, but only enables it when BR_WITH_QTNETWORK is on, so using qtnetwork can remain off by default. 

@jklontz, @sklum  care to review?

### DIFF
--- a/openbr/openbr.cpp
+++ b/openbr/openbr.cpp
@@ -318,7 +318,7 @@ const char *br_version()
 
 void br_slave_process(const char *baseName)
 {
-#ifndef BR_EMBEDDED
+#ifdef BR_WITH_QTNETWORK
     WorkerProcess *worker = new WorkerProcess;
     worker->transform = Globals->algorithm;
     worker->baseName = baseName;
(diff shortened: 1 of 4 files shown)


openbr/openbr.cpp
```diff
@@ -316,6 +316,20 @@ const char *br_version()
     return version.data();
 }
 
+void br_slave_process(const char *baseName)
+{
+#ifdef BR_WITH_QTNETWORK
```
jklontz: Does this also need to check BR_EMBEDDED? Or does BR_EMBEDDED=ON trigger BR_WITH_QTNETWORK=OFF?

caotto: BR_WITH_QTNETWORK  is off by default, so someone would have to go out of
their way to build br_embedded and also set BR_WITH_QTNETWORK=ON.

As far as I understand the check on br_embedded was even added in the first
place is because of the use of qtnetwork in processwrapper, so probably
this is just a more specific check.

jklontz:
Thanks!

Issue #468, opened 2016-04-15
After build -algorithm Fatal error


DadSimen:
Hello,
when i was trying tutorial first line for opening webcam works perfectly. Second example give me error:
```

C:\Users\JIO>br -gui -algorithm "Cvt(Gray)+Show(false)" -enroll 0.webcam
Set algorithm to Cvt(Gray)+Show(false)
Fatal: Failed to set br::Transform* br::IndependentTransform::transform to:
  SDK Path: C:/openbr
  File:
  Function:
  Line: 0
```

Any sugestions ?

d1xlord:
Hi, having this same issue. Any solutions regarding this?

DadSimen:
Hello,

I am trying to make it working but no success here. I find out that problem is probably in Qt. I am using wrong version of Qt 5.6 because of Windows 10 and i should use Qt 5.4 but there is no support for win10. Right now i am trying to use Qt 5.5 – i found somewhere build of Qt5.5 for win10 but still no success.

cristobaldobranco:
I'm seeing the same issue with Qt 5.6 for Android in a compilation of OpenBR for Android. For me it was possible to switch to Qt 5.5.1 for Android and get over this stage.

DepthDeluxe:
I've had issues with Qt 5.6 on Arch Linux x64.  I think there is some sort of change in API that is causing the problem.  Switching to Qt 5.5.1 solved the problem

sklum:
I'm going to try to solve this issue when I get some free time.  However, if anyone has made local modifications to openbr to resolve compatibility with Qt 5.6 we would welcome them in a PR.

d1xlord:
I also got over this issue by switching to Qt 5.4

Thanks for the reply

raomshauty:
Hi -- I ran into this with compile openbr against qt 5.7.  I spent time looking for solution and came up with this change -- not super sure if its the right fix because I cannot understand exactly why its a problem with new qt version but not with old one ... but you can try and see if its working for you.

In method Transform::make within openbr_plugin.cpp -- change usage of qVariantFromValue<void*>(transform) to QVariant::fromValue(transform).

old
```
// causes QObject::setProperty to return false inside Object::setProperty  
        independent.set("transform", qVariantFromValue<void*>(transform));
```

new
```
// causes QObject::setProperty to return true as expected -- seems to work
        independent.set("transform", QVariant::fromValue(transform));
```

[...]

PR #486, opened 2016-07-05
Possible fix for run-time crash when using qt 5.7

raomshauty:
Possible fix for runtime crash when using qt 5.7.  Without this fix -- crash happens because QObject::setProperty(qPrintable(name), value) returns false within Object::setProperty


raomshauty:
Fix  #468

jklontz:
Thanks! I'm going to merge this under the assumption that it's backwards compatible with earlier versions of Qt 5. If we run into issues we'll have to revert. @sklum heads up.

Issue #488, opened 2016-07-11
Gallery Append


sklum:
@caotto When you have some free time, could you fix the Gallery `append` option?  I believe it was broken when we switched to using `StreamTransform` for enrollment.  At the moment, based on my understanding, adding the `append` option only skips writing a template to the gallery that's already in the gallery whereas I'd like it to skip passing the a template through the given algorithm.

caotto:
So, the problem isn't functional but efficiency thing (i.e. you want to avoid doing processing on images already present in gallery?) 

It should be pretty easy, this is currently handled here:
https://github.com/biometrics/openbr/blob/master/openbr/core/core.cpp#L212
where FileExclusion is the thing that removes overlap filenames.

It's currently placed at the endpoint, but it should be as simple as making a FileExclusion transform, and combining it with the normal pipeline (at the beginning) with 
https://github.com/biometrics/openbr/blob/master/openbr/openbr_plugin.h#L940

Anyway, I can do it but don't have time till a little later this week

sklum:
Exactly.  Glad to hear it's a simple fix - no rush!

sklum:
Done!

PR #489, opened 2016-07-14
Adjust enroll append behavior

caotto:
per discussion in #488, adjust behavior of "append" flat during enrollment to exclude files at the beginning rather than end of processing

@sklum care to review? This appears to work for me based on some brief testing on MEDS data, but you should check for your usecase.


sklum:
Just getting to this - tested and looks good.  Thanks @caotto.

PR #490, opened 2016-07-17
CUDA-accelerated PCA training and improved PCA projection speeds

DepthDeluxe:
I used the NVIDIA cuBLAS and cuSOLVER libraries to implement PCA training following the same approach that was taken for CPU PCA training.  The covariance matrix is computed, then I used SVD to pull the eigenvectors and eigenvalues out of the system.

I have performed memory checking via `cuda-memcheck` on the ATT face database at both 96x96 and 128x128 in size and it has returned no memory errors or memory leaks.

I have also analyzed the differences and attached in to the PR.  Take a look at [comparison.csv.zip](https://github.com/biometrics/openbr/files/367614/comparison.csv.zip).

PCA projection is now ~50% faster than the previous CUDA implementation because it uses the cuBLAS matrix-vector multiplication function.


openbr/plugins/cuda/util.cpp
```diff
@@ -0,0 +1,62 @@
+#include <iostream>
```
jklontz: Is there a missing `util.h` file?

DepthDeluxe: hmm... these functions are entirely meant for debugging.  I compiled them in so I could print out dumps through GDB.  Do you think it would be a better idea to add a `util.h` or remove the whole file from the project?

jklontz: Remove I think, `core/eigenutils.h` provides similar functionality already.

DepthDeluxe:
@jklontz removed `util.cpp`

jklontz:
Thanks!

PR #491, opened 2016-07-19
added timing report for training with StopWatch

DepthDeluxe:
StopWatch currently only reports the projection time.  I modified the MetaTransform to print out the same report for training.

### DIFF
--- a/openbr/plugins/metadata/stopwatch.cpp
+++ b/openbr/plugins/metadata/stopwatch.cpp
@@ -17,10 +17,9 @@
 #include <opencv2/opencv.hpp>
 using namespace cv;
 
-#include <openbr/plugins/openbr_internal.h>
+#include <QAtomicInt>
 
-#include <iostream>
-using namespace std;
+#include <openbr/plugins/openbr_internal.h>
 
 namespace br
 {
@@ -38,10 +37,9 @@ class StopWatchTransform : public MetaTransform
     BR_PROPERTY(QString, description, "Identity")
 
     br::Transform *transform;
-    mutable QMutex mutex;
-    mutable long milliseconds;
-    mutable long images;
-    mutable long pixels;
+    mutable QAtomicInt milliseconds;
+    mutable QAtomicInt images;
+    mutable QAtomicInt pixels;
 
 public:
     StopWatchTransform()
@@ -52,9 +50,9 @@ public:
 private:
     void reset()
     {
-        milliseconds = 0;
-        images = 0;
-        pixels = 0;
+        milliseconds.fetchAndStoreOrdered(0);
+        images.fetchAndStoreOrdered(0);
+        pixels.fetchAndStoreOrdered(0);
     }
 
     void init()
@@ -69,13 +67,16 @@ private:
 
         transform->train(data);
 
-        milliseconds += watch.elapsed();
-        images += data.size();
+        milliseconds.fetchAndAddOrdered((int)watch.elapsed());
+        images.fetchAndAddOrdered(data.size());
 
         // compute the total number of pixels we processed
         foreach(const TemplateList &list, data) {
-          const Mat& m = list.first().m();
-          pixels += m.rows*m.cols;
+          foreach(const Template &t, list) {
+            foreach(const cv::Mat &m, t) {
+              pixels.fetchAndAddOrdered(m.rows*m.cols);
+            }
+          }
         }
     }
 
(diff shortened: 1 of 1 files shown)


openbr/plugins/metadata/stopwatch.cpp
```diff
@@ -14,8 +14,14 @@
  * limitations under the License.                                            *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+#include <opencv2/opencv.hpp>
+using namespace cv;
+
 #include <openbr/plugins/openbr_internal.h>
 
+#include <iostream>
+using namespace std;
```
sklum: Is this header required?

openbr/plugins/metadata/stopwatch.cpp
```diff
@@ -58,7 +64,19 @@ class StopWatchTransform : public MetaTransform
...
+        watch.start();
+
         transform->train(data);
+
+        milliseconds += watch.elapsed();
+        images += data.size();
+
+        // compute the total number of pixels we processed
+        foreach(const TemplateList &list, data) {
+          const Mat& m = list.first().m();
```
sklum: Are templates here guaranteed to have only a single matrix?

sklum:
@DepthDeluxe A few minor comments - looks great otherwise.  Keep these awesome commits up!

openbr/plugins/metadata/stopwatch.cpp
```diff
@@ -33,7 +39,7 @@ class StopWatchTransform : public MetaTransform
 
     br::Transform *transform;
     mutable QMutex mutex;
-    mutable long miliseconds;
+    mutable long milliseconds;
```
sklum: One other quick note, you might change these to `QAtomicInt` to avoid the need for a lock in projection and to make sure updating this value is safe in the newly added training code.

openbr/plugins/metadata/stopwatch.cpp
DepthDeluxe: I spent a decent amount of time trying to understand what was going on, from my experimentation, if you are training N images, the input `data` consisted of N `TemplateList`s each with a single image in them.  I don't know why the data was passed this way, seems very inefficient.  Actually, could this potentially be different if you are passing video in?

openbr/plugins/metadata/stopwatch.cpp
jklontz: ^

openbr/plugins/metadata/stopwatch.cpp
caotto: It's basically for maintaining consistency with how project works at test time.

Functionally, at this point everything is actually done in terms of project(templatelist,templatelist), and transforms can change the size of those lists (well, should only do so if -enrollAll true is set). So e.g. a detector can receive a template (corresponding to one image) as input, and output multiple templates (if multiple faces are detected), or no templates (if no faces are detected). Or various ways to handle video are also possible through this system (e.g. you could sample every nth frame or something), but I'm not aware that anyone has done much actual work with that. 

So, the size of the lists could actually change, but pretty much only if train is done with -enrollAll set to true. Actually I'm not sure this is really worth maintaining for training, typically people will just use pre-defined labeled detections or pre-cropped images--but I don't know that it is a noticeable performance issue (as far as I'm aware, training time is typically dominated by interiors of train methods doing linear algebra or optimization problems).

sklum: @DepthDeluxe As @caotto has indicated, it's theoretically possible to have multiple matrices per `Template` in the `TemplateList`.  It's a bit of an edge case to handle, but just to make sure we're covering all of our bases here.  You'll note that the `project` function iterates over all matrices in the `src` `Template`, so you could just do things the same way within this loop.

openbr/plugins/metadata/stopwatch.cpp
```diff
@@ -46,9 +50,9 @@ class StopWatchTransform : public MetaTransform
 private:
     void reset()
     {
-        miliseconds = 0;
-        images = 0;
-        pixels = 0;
+        milliseconds.fetchAndStoreOrdered(0);
```
DepthDeluxe: @sklum not too sure about the lockless threading stuff so I erred on the conservative side using "Ordered" memory preference.  Is this the safe and correct way to be setting a `QAtomicInt`?

jklontz:
@sklum Since you've been reviewing, can you give the final approval and merge? Thanks!

openbr/plugins/metadata/stopwatch.cpp
sklum: @DepthDeluxe This looks like the correct way of doing things based on my understanding of `QAtomicInt`.  Thanks.

sklum:
@DepthDeluxe Sorry for the delay - missed the commits yesterday.  Looks great!  Thanks again.

PR #496, opened 2016-08-23
Modifed Expand Rect to allow the use of propName

Keyurpatel93:


openbr/plugins/metadata/expandrect.cpp
```diff
@@ -28,16 +28,24 @@ namespace br
 class ExpandRectTransform : public UntrainableTransform
 {
     Q_OBJECT
+    Q_PROPERTY(QString propName READ get_propName WRITE set_propName RESET reset_propName STORED false)
```
jklontz: @Keyurpatel93 new parameters are generally added at the end not the beginning as this changes the API. What was `ExpandRect(.75,1.2)` is now `ExpandRect(,.75,1.2)` in this code.

PR #502, opened 2016-10-04
Updated empty gallery to use deepest folder as subject label

sklum:
@jklontz As discussed.

### DIFF
--- a/openbr/core/utility.cpp
+++ b/openbr/core/utility.cpp
@@ -19,6 +19,25 @@ QStringList br::getFiles(QDir dir, bool recursive)
     return files;
 }
 
+QList<QPair<QString,QStringList> > br::getFilesWithLabels(QDir dir)
+{
+    dir = QDir(dir.canonicalPath());
+
+    QStringList files;
+    foreach (const QString &file, QtUtils::naturalSort(dir.entryList(QDir::Files)))
+        files.append(dir.absoluteFilePath(file));
+
+    QList<QPair<QString,QStringList> > filesWithLabels;
+    filesWithLabels.append(QPair<QString,QStringList>(dir.dirName(),files));
+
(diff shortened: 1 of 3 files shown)


openbr/core/utility.cpp
```diff
@@ -19,6 +19,25 @@ QStringList br::getFiles(QDir dir, bool recursive)
     return files;
 }
 
+QList<QPair<QString,QStringList> > br::getFilesWithLabels(QDir dir)
+{
+    dir = QDir(dir.canonicalPath());
+
+    QStringList files;
+    foreach (const QString &file, QtUtils::naturalSort(dir.entryList(QDir::Files)))
```
jklontz: You should be able to drop the `QtUtils::naturalSort` in both uses.

openbr/plugins/gallery/empty.cpp
```diff
@@ -113,10 +115,11 @@ class EmptyGallery : public Gallery
 
     static TemplateList getTemplates(const QDir &dir)
     {
-        const QStringList files = getFiles(dir, true);
-        TemplateList templates; templates.reserve(files.size());
-        foreach (const QString &file, files)
-            templates.append(File(file, dir.dirName()));
+        const QList<FilesWithLabel> filesWithLabels = getFilesWithLabels(dir);
+        TemplateList templates; templates.reserve(filesWithLabels.size());
```
jklontz: Drop the reserve, since it's the subject count instead of the file count it no longer makes much sense.

openbr/plugins/gallery/empty.cpp
```diff
@@ -20,6 +20,8 @@
 #include <openbr/core/qtutils.h>
 #include <openbr/core/utility.h>
 
+typedef QPair<QString,QStringList> FilesWithLabel;
```
jklontz: Move this typedef into `utility.h` and use it everywhere?

jklontz:
Consider my feedback then merge when ready, thanks!

PR #504, opened 2016-10-13
basic infrastructure for handling rotated rects

jklontz:


jklontz:
@sklum said it looks good.

PR #505, opened 2016-10-13
rotate should preserve center and size

jklontz:


jklontz:
@sklum says it looks good.

PR #507, opened 2016-10-19
Only use old metadata for an empty file

sklum:
@jklontz This addresses the issue I was alluding to earlier.  Any problems?


jklontz:
I'm not convinced this is a good idea. Can you give me an example of where this was useful?

sklum:
If you try to enroll a set of files as `br -algorithm "FUBAR" -enroll test1.gal[param1=0] test.gal[param2=1]`, without this patch, `test1.gal` is set to have `param2=1` because of the way splitting is done.  This is fixing a bug, not related to usefulness.

jklontz:
oh, k. I had read it as param1=0 would also be assigned to test.gal. looks good then

PR #510, opened 2016-10-27
Added flag to random rects to allow for the random rects to be of the…

Keyurpatel93:
… same size and ensure the patches don't contain black pixels


bklare:
@Keyurpatel93 Looks good. Thanks!

PR #515, opened 2016-12-02
miscellaneous scripts 

Keyurpatel93:
…keys exist in the templates metadata

roi.cpp copyOnCrop=True is now the default
rotate.cpp rotates a mat by 90 intervals
extractvaluefromLlist.cpp extracts a single value from a list in metadata and saves it in a new key
majorityvoting.cpp Calculates which class has the highest scores when given a list of scores for each class
scorelevelfusion.cpp Fuses the scores together for each class when given a list of scores for each class

### DIFF
--- a/openbr/plugins/classification/caffe.cpp
+++ b/openbr/plugins/classification/caffe.cpp
@@ -2,7 +2,7 @@
 #include <openbr/core/opencvutils.h>
 
 #include <caffe/caffe.hpp>
-#include <caffe/layers/memory_data_layer.hpp>
+//#include <caffe/layers/memory_data_layer.hpp>
 
 using caffe::Caffe;
 using caffe::Net;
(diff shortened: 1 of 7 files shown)


openbr/plugins/classification/caffe.cpp
```diff
@@ -2,7 +2,7 @@
 #include <openbr/core/opencvutils.h>
 
 #include <caffe/caffe.hpp>
-#include <caffe/layers/memory_data_layer.hpp>
+//#include <caffe/layers/memory_data_layer.hpp>
```
jklontz: This change shouldn't be comitted

jklontz:
LGTM otherwise

Keyurpatel93:
Uncommented out specified line https://github.com/biometrics/openbr/pull/515/commits/dc67cf8cff88aa217be2e21f2c23b240b042d003

jklontz:
Thanks!

PR #527, opened 2017-01-30
Created new transform checkrects.cpp

Keyurpatel93:


sklum:
Extra points if you can merge this functionality with that in `FixRects`.

Keyurpatel93:
@jklontz I didn't inherit from UntrainableMetadataTransform because the checkRects transform needs to operate on templates and not files as I need the Mat dimensions. I figured that @sklum  also didn't inherit his FixRectsTransform due to this same problem.

jklontz:
Oh yeah, good point!

PR #579, opened 2021-01-25
Changes to evalEER

JordanCheney:
This PR makes 2 changes to evalEER and adds a new, corresponding, plotEER method. The C API and command line tool are updated to support the new function.

Changes are:

1. Change how operating points are computed to mirror eval (i.e. operating points are added only if both the true positive AND false positive count changes). This fixes a bug where successive operating points could have the same FAR, which caused nan results in the output.

2. Change the optional pdf argument to a CSV argument (again mirroring eval) so that results are written as CSVs and can be combined to show multiple results on the same plot


bhklein:
looks good!

PR #580, opened 2021-02-26
Opencv4

JordanCheney:
This commit updates OpenBR to be compatible with OpenCV 4. There were a few challenges with this-

1. OpenCV 4 removed the C persistence layer and the C ML classes, which the OpenBR boost interface depends on. I solved this by pulling the relevant classes out of OpenCV 2.X and 3.X and compiling them as part of the OpenBR build. See `3rdparty/opencv2-compat`
2. The OpenCV Random Forest API changed significantly, to the point where the `plugins/classification/forest.cpp` functionality is significantly impacted. In this patch `forest.cpp` is excluded from the build
3. The OpenCV keypoint detector API changed significantly, to the point where `plugins/imgproc/keypointdescriptor.cpp` and `plugins/metadata/keypointdetector.cpp` are significantly impacted. Someone with interest in those plugins would need to rethink the interface to the `cv::Features2D` interface. In this patch both plugins are excluded from the build
4. The plugin `plugins/imgproc/custom_sift.cpp` relied on function definitions that were no longer available. A similar approach to 1) might be possible (i.e. find the function definitions in a previous version of OpenCV and bring the source into OpenBR) but I couldn't find those functions. In this path `custom_sift.cpp` is excluded from the build


sklum (commented):
No major issues here, but a question the boost change.

PR #588, opened 2021-09-01
Obsolete plugins

jklontz:
Remove plugins that are no longer relevant in the modern deep learning era.


JordanCheney:
I'd be in favor of keeping DLib as OpenBR doesn't have another landmarker but otherwise this list is comprehensive

bhklein (approved):

PR #595, opened 2023-07-18
opencv4 bug fixes

Keyurpatel93:


openbr/plugins/classification/svm.cpp
```diff
@@ -79,8 +79,8 @@ class SVMTransform : public Transform
     Ptr<ml::SVM> svm;
 
 private:
-    BR_PROPERTY(Kernel, kernel, Linear)
-    BR_PROPERTY(Type, type, C_SVC)
+    BR_PROPERTY(Kernel, kernel, RBF)
```
JordanCheney, line 82: It might not matter, but changing the defaults is a breaking change for any other scripts that use `SVM` and expect a Linear kernel. It might be better to just change the arguments at your call site to `RBF`/`NU_SVR`

JordanCheney (changes requested):

PR #599, opened 2023-11-10
Inplace eval

bhklein:
Simplified `InplaceEval` that makes use of a pre-computed mask rather than target and query files.

### DIFF
--- a/openbr/core/eval.cpp
+++ b/openbr/core/eval.cpp
@@ -491,13 +491,8 @@ void openAndReadMatrixHeader(QFile &file, size_t &rows, size_t &cols, size_t &by
 // Helper struct for statistics accumulation
 struct Counter
 {
-    float truePositive, falsePositive, falseNegative;
-    Counter()
-    {
-        truePositive = 0;
-        falsePositive = 0;
-        falseNegative = 0;
-    }
+    double truePositive, falsePositive, falseNegative;
+    Counter() : truePositive(0), falsePositive(0), falseNegative(0) {}
 };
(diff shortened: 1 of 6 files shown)


openbr/core/eval.cpp
```diff
@@ -474,233 +469,197 @@ void assertEval(const QString &simmat, const QString &mask, float accuracy)
...
+    size_t sdPoints = qMin(qMin((size_t)Max_Points, genuineCount), impostorCount);
+    size_t genuineRate = genuineCount / sdPoints;
+    size_t impostorRate = impostorCount / sdPoints;
+    size_t genuinesSampled(0), sampleNextGenuine(0), impostorsSampled(0), sampleNextImpostor(0);
+
+    size_t fps(0), tps(0);
+    QList<float> thresholds = stats.keys();
+    std::sort(thresholds.begin(), thresholds.end());
+    std::reverse(thresholds.begin(), thresholds.end());
+    QList<OperatingPoint> operatingPoints;
```
bhklein, line 572: This is just for convenience of printing out operating points at exact FAR or FRR values.

openbr/core/eval.cpp
```diff
@@ -474,233 +469,197 @@ void assertEval(const QString &simmat, const QString &mask, float accuracy)
...
+    for (int i=0; i<thresholds.size(); i++) {
```
bhklein, line 573: We log every operating point now.

Keyurpatel93:
looks good to me

PR #602, opened 2024-02-02
Add correlation calculation to evalRegression.

JStehouwer:


openbr/core/eval.cpp
```diff
@@ -1090,9 +1090,25 @@ float EvalLandmarking(const QString &predictedGallery, const QString &truthGalle
...
+// See Pearson calculation as utilized by python numpy.
+struct CorrelationCounter
+{
+    double sumX, sumY, sumXX, sumYY, sumXY;
+    CorrelationCounter() : sumX(0.0), sumY(0.0), sumXX(0.0), sumYY(0.0), sumXY(0.0) {}
+
+    double correlation_coefficient(int length)
+    {
+        double numer = (length * sumXY) - (sumX * sumY);
+        double denomX = (length * sumXX) - (sumX * sumX);
```
Keyurpatel93, line 1103: shouldn't `sumX * sumX` be `sumXX`. Although I believe both would have the same value regardless.

PR #620, opened 2025-03-04
Nixed --object-directory param from gcovr as it limits search_path in…

NateK-ROC:
…trinsics to only the specified directory


KOConchobhair (approved):
LGTM

PR #641, opened 2025-12-15
Openbr docs app

JStehouwer:


JStehouwer:
This adds `docs(...)` functionality to OpenBR transforms which are exposed via two endpoints.

Print out docs for a specific algorithm string:
```
$ br -algorithm "Read+Fork([Blur,Resize,Discard+Read])" -docs
Set algorithm to Read+Fork([Blur,Resize,Discard+Read])
Pipe(Transform transform): Project the output of each Transform into the next
    Read(Mode mode): Read the image from disk
    Fork(Transform[] transforms): Project the input template(s) through each transform and concatenate the results
        .BlurTransform->docs() not implemented
        .ResizeTransform->docs() not implemented
        Pipe(Transform transform): Project the output of each Transform into the next
            Discard(): Removes all matrices from a template.
            Read(Mode mode): Read the image from disk
```

Print out docs for all transforms (optional regex name matching):
```
$ br-docs --regex "Cvt"   
...
    .CvtTransform->docs() not implemented
    CvtFloat(): Convert the matrix to float32
    CvtUChar(): Convert the matrix to uint8
```

JStehouwer:
TODO - Resolve issues when generating docs for transforms when the `init()` function fails when using default arguments.

openbr/plugins/metadata/eyes.cpp
```diff
@@ -71,7 +71,8 @@ class ASEFEyesTransform : public UntrainableTransform
 
         // Open the eye locator model
         file.setFileName(Globals->sdkPath + "/share/openbr/models/EyeLocatorASEF128x128.fel");
-        if (!file.open(QFile::ReadOnly)) qFatal("Failed to open %s for reading.", qPrintable(file.fileName()));
+        if (!file.open(QFile::ReadOnly))
+            throw QString("Failed to open %1 for reading.").arg(file.fileName()).toStdString();
```
jklontz, line 75: Got `br -docs` working again. How about this approach for handling initialization errors? I think you can remove _init()?

app/br-docs/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+add_executable(br-docs br-docs.cpp ${BR_RESOURCES})
```
jklontz, line 2: I don't think we need a separate CLI app here. I think `br -docs` should do this when `-algorithm` isn't set.

app/br/br.cpp
```diff
@@ -210,6 +210,9 @@ class FakeMain : public QRunnable
             else if (!strcmp(fun, "help")) {
                 check(parc == 0, "No parameters expected for 'help'.");
                 help();
+            } else if (!strcmp(fun, "docs")) {
```
jklontz, line 213: delete app/br-docs now?

openbr/openbr.h
```diff
@@ -26,6 +26,8 @@ extern "C" {
 
 BR_EXPORT const char *br_about();
 
+BR_EXPORT void br_docs();
```
jklontz: Would prefer to design this function like br_about() and return a string. The caller then has flexibility to choose between printf and other options.

openbr/openbr_plugin.h
```diff
@@ -752,6 +753,29 @@ class BR_EXPORT Transform : public Object
     virtual void train(const TemplateList &data);
     virtual void train(const QList<TemplateList> &data);
 
+    void print_doc_header(QString doc, int indent) const {
+        QString name = this->file.suffix();
```
jklontz: Move the implementation of this non-trival function into openbr_plugin.cpp

JStehouwer: Resolved in [62d5a3d](https://github.com/biometrics/openbr/pull/641/commits/62d5a3d9dd161593adfb2dc39dc477ac45645669).

openbr/plugins/metadata/eyes.cpp
JStehouwer: Resolved by using exceptions in transform `init()` calls instead of `qFatal()`.

app/br/br.cpp
JStehouwer: Discussed in person, keeping `br-docs` due to its additional functionality (regex filtering, etc).

openbr/openbr.h
JStehouwer: This should be resolved by [1a28430](https://github.com/biometrics/openbr/pull/641/commits/1a2843004cae51ecbedcc5e4eb18aadae1109c74).

JStehouwer:
This idea is being dropped as it's unlikely to be utilized. Closing

